PR #117, opened 2026-06-11
feat(cron): durable in-flight marker + startup recovery for interrupted jobs (issue 105)

### LINKED ISSUE #105: [CAPABILITY] Cron agent sessions do not survive gateway restarts — interrupted jobs vanish without record or retry
## Problem (observed 2026-06-10, cost a full daily cycle)
`evolution-implementation` fired on schedule at 22:00:33 and worked for ~30 minutes (46 API calls). At 22:30:47 the gateway was restarted (operator was applying provider-failover config via Telegram; the agent itself initiated the planned restart). The cron session was killed mid-flight:
- no result was recorded — `hermes cron list` shows NO "Last run" for the job at all (as if it never ran);
- no retry / no re-fire — the daily slot was simply lost;
- downstream `evolution-integration` ran at 23:01, found nothing, reported "ok" — the whole night looked healthy while producing nothing.

Session log ends abruptly: `agent.log` last entry for `cron_ac158635a029_20260610_220033` at 22:30:32, then the new gateway process initializes MCP at 22:30:51.

## Why it matters
Gateway restarts are ROUTINE (config changes, `hermes update` auto-update at 04:17 nightly, /restart). Any cron job unlucky enough to overlap one silently dies. Combined with the lack of failure records this is invisible — #83 (watchdog) would detect the gap next day, but the work is still lost.

## Proposed direction
Any of (in increasing ambition):
1. On gateway startup, detect cron sessions that were in-flight at shutdown (marker file written at job start, cleared at completion) and record them as `interrupted` — making the failure visible to `cron list` and the future watchdog (#83).
2. Re-fire interrupted jobs once on startup if still within N hours of their scheduled slot.
3. Graceful drain: planned restarts already drain user sessions — extend the drain to wait (bounded) for running cron sessions, or delay the restart until the job completes.

## Value
- Impact: 0.8 (a routine operation silently destroys daily cycles; happened on day one of observation)
- Effort: 0.4 (option 1 is a marker file + startup scan; 2-3 incremental)
- Priority Score: 4.0

Lexus2016:
Implements issue 105 (operator-implemented per owner's request).

## Problem
A routine gateway restart (config change, nightly auto-update, /restart) kills any in-flight cron job with **no record at all**: no last_run, no error, no retry. The 2026-06-10 nightly cycle was lost exactly this way — `evolution-implementation` died 30 minutes into its run, `hermes cron list` showed no trace it ever fired, and downstream `integration` reported ok-on-nothing.

## Design (respects existing at-most-once semantics)
- **`mark_job_started()`** — durable `state='running'` + `run_started_at` in jobs.json at job start; `mark_job_run()` already flips state back on completion.
- **`recover_interrupted_jobs(max_refire_age_hours=6)`** — runs ONCE at gateway startup before the first tick (nothing runs in the new process yet ⇒ every persisted `running` marker is stale). Records `last_status='interrupted'` + explanatory `last_error`, restores `state='scheduled'`, and **re-fires only recent deaths** (`next_run_at=now`). The bounded window preserves `advance_next_run()`'s deliberate anti-crash-loop design — old markers just wait for their normal slot.
- Deliberately NOT called from CLI `cron tick`: a tick alongside a live gateway would misread its in-flight jobs as interrupted.
- Composes with the watchdog (#116): it independently alerts on jobs stuck `running` >12h.

## Tests
7 new (marker set/clear, recent-refire, old-no-refire, healthy untouched, completed-not-interrupted); full `tests/cron/` 417/417 locally.

Closes #105

### DIFF
--- a/cron/jobs.py
+++ b/cron/jobs.py
@@ -1009,6 +1009,84 @@ def advance_next_run(job_id: str) -> bool:
         return False
 
 
+def mark_job_started(job_id: str) -> None:
+    """Record that a job's execution has begun (interrupted-run forensics).
+
+    Leaves a durable ``state="running"`` + ``run_started_at`` marker in
+    jobs.json. ``mark_job_run()`` clears it on completion (it sets state to
+    scheduled/completed/error). If the gateway dies mid-run — a routine
+    planned restart is enough — the marker survives on disk, and
+    ``recover_interrupted_jobs()`` turns it into a visible ``interrupted``
+    record on the next startup instead of the run silently vanishing
+    (issue 105: the 2026-06-10 nightly cycle was lost exactly this way).
+    """
+    with _jobs_file_lock:
+        jobs = load_jobs()
+        for job in jobs:
+            if job["id"] == job_id:
+                job["state"] = "running"
+                job["run_started_at"] = _hermes_now().isoformat()
+                save_jobs(jobs)
+                return
+        logger.warning("mark_job_started: job_id %s not found", job_id)
+
+
+def recover_interrupted_jobs(max_refire_age_hours: float = 6.0) -> List[str]:
+    """Turn stale ``running`` markers into visible interrupted records.
+
+    Call ONCE at gateway startup, BEFORE the first scheduler tick: this
+    process has not started any job yet, so every job still marked
+    ``running`` was killed mid-run by the previous process's death.
+
+    For each such job: record ``last_status="interrupted"`` with an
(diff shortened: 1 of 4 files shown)


Lexus2016:
Re-opened by realized-impact gate: post-merge verdict was regressed (2026-09-02) — the fix did not reduce the target signal (per-session rate rose >20% over baseline). Needs a focused regression before re-closing.

PR #126, opened 2026-06-11
feat: STT local fallback when cloud provider unavailable (Closes #122)

### LINKED ISSUE #122: [CAPABILITY] STT provider missing — voice messages in Telegram are silently dropped
The agent repeatedly cannot process voice messages sent by users in Telegram sessions because no STT (Speech-to-Text) provider is configured. Every time a user sends a voice message, the agent responds with a boilerplate apology and instructions, but never actually transcribes the content, leaving the user's request unfulfilled.

- Frequency: seen in 16 of 24 recent sessions (67% of sessions in the last 7 days)
- Tool/area involved: Telegram message reception / STT subsystem
- Failure shape: "[The user sent a voice message but I can't listen to it right now — no STT provider is configured.]" — message is silently dropped, user must re-type request in text

### Impact on real tasks
Voice messages are a primary input modality for Telegram users. When the agent cannot process them, the user must re-type everything, doubling interaction cost and frequently causing task abandonment. In group chats this is especially disruptive because other participants see repeated failed attempts.

### Proposed direction
Add STT provider integration (e.g., Whisper API, local whisper.cpp, or Telegram's built-in voice-to-text) so voice messages are transcribed automatically before the agent processes them. The transcription should be injected into the user turn as text content.

### Value
- Impact: 1.0 (task impossible for voice-only input)
- Effort: 0.3 (standard integration)
- Priority Score: 6.67

Lexus2016:
Automated evolution PR for issue #122.

### DIFF
--- a/tests/gateway/test_stt_config.py
+++ b/tests/gateway/test_stt_config.py
@@ -16,7 +16,9 @@ def test_gateway_config_stt_disabled_from_dict_nested():
     assert config.stt_enabled is False
 
 
-def test_load_gateway_config_bridges_stt_enabled_from_config_yaml(tmp_path, monkeypatch):
+def test_load_gateway_config_bridges_stt_enabled_from_config_yaml(
+    tmp_path, monkeypatch
+):
     hermes_home = tmp_path / ".hermes"
     hermes_home.mkdir()
     (hermes_home / "config.yaml").write_text(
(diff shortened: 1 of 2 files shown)


Lexus2016:
Closing this PR: integration code-review found a real, deterministic bug (not a flaky test). `test (4)` fails every run:

`test_no_provider_returns_error` → `AssertionError: "No STT provider" in "Local transcription failed: not enough values to unpack (expected 2, got 0)"`

Root cause: the new last-resort local fallback routes `provider=="none"` into `_transcribe_local`, which crashes on `segments, info = _local_model.transcribe(...)` when no usable model is loaded, and it also changed the no-provider contract without updating the existing test.

The idea (issue #122 — local STT fallback) is still wanted; the implementation needs rework. A specific rework brief is posted on #122, which is flipped back to `needs-work`. Re-open a fresh PR once the no-provider path returns a clean error and the fallback never crashes on unpack. Branch kept for reference.

PR #127, opened 2026-06-11
fix: Allow U+200D ZWJ in multilingual scripts (Arabic, Cyrillic, Devanagari, etc.) (Closes #123)

### LINKED ISSUE #123: [FIX] Cron injection scanner false-positives on harmless invisible unicode (U+200D) in multilingual prompts
Cronjobs with assembled prompts are repeatedly blocked by the injection scanner claiming "invisible unicode" (e.g., U+200D Zero Width Joiner). This is a false positive — the invisible characters are legitimate parts of Cyrillic text or Telegram formatting, not adversarial injections. The cronjob is silently aborted, and the user only sees a failure notification in Telegram.

- Frequency: seen in 6 of 24 recent sessions (25%)
- Tool/area involved: Cron scheduler / prompt injection scanner
- Failure shape: `blocked: prompt contains invisible unicode u+200d (possible injection)` — cron output never generated

### Impact on real tasks
Automated daily digests and reports that contain Cyrillic content or emoji are killed before they reach the model. Users depend on these cronjobs for channel content; the false-positive rate makes the digest pipeline unreliable.

### Proposed direction
Tune the injection scanner so that common harmless invisible characters (ZWJ, ZWNJ, directional marks in multilingual scripts) are not blanket-blocked in cron contexts. Alternatively, add a `strict_injection_scan: false` flag for trusted cronjobs, or normalize the prompt before scanning.

### Value
- Impact: 0.8 (task fails silently, user only sees broken digest)
- Effort: 0.3 (scanner rule adjustment)
- Priority Score: 5.33

Lexus2016:
Automated evolution PR for issue #123.

### DIFF
--- a/tests/cron/test_cron_prompt_injection_skill.py
+++ b/tests/cron/test_cron_prompt_injection_skill.py
@@ -450,3 +450,38 @@ class TestScriptOutputNotStrictScanned:
         }
         with pytest.raises(scheduler.CronPromptInjectionBlocked):
             scheduler._build_job_prompt(job)
+
+
+
+class TestInvisibleUnicodeMultilingualZWJ:
+    """U+200D Zero-Width Joiner in multilingual scripts must not trigger a
+    hard block. Only ZWJ between plain Latin/digits/punctuation should
+    remain blocked (injection-hiding)."""
+
+    def test_cyrillic_zwj_allowed(self, cron_env):
+        _, scheduler = cron_env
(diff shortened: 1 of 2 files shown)


Lexus2016:
Re-opened by realized-impact gate: post-merge verdict was regressed (2026-09-02) — the fix did not reduce the target signal (per-session rate rose >20% over baseline). Needs a focused regression before re-closing.

Issue #304, opened 2026-06-17
[CHILD/#226] Agent-as-Judge — trace replay subsystem


persik-77:
Child of #226 (decomposition — parent labelled `needs-decomposition`).

Builds on agent/agent_judge.py (already shipped). Add a replay layer: given a recorded session, replay the key steps (model decisions, tool calls) deterministically and let the judge score each step rather than only the final output. Effort: 0.5.

**Parent**: #226 (decomposition — parent should be labelled `needs-decomposition`)

_This issue is one shippable slice of its parent's decomposition. Effort is bounded to <= 0.6 so it can compete normally in the priority cycle._

Lexus2016:
Implemented first increment in #307 (trace replay subsystem).

Lexus2016:
Blocked: CI failing on PR #307 (test slice 5 and nix/ubuntu-latest). Before re-opening a PR:

1. Investigate and fix the root cause of the test(5) failure in this branch — the failures are not present on main for the same commit base, so they are introduced by the replay changes.
2. Verify the nix (ubuntu-latest) workflow file is valid and the job can be re-run from the GitHub UI; if the workflow is corrupted in this branch, fix it.
3. Confirm every new top-level symbol (replay_trace_steps, score_replayed_trace) is imported/called from a real non-test module in a later increment, or wire dashboard/CLI integration in this increment if you want it mergeable now.

Re-open a PR once CI is fully green and re-runnable.

PR #307, opened 2026-06-17
feat: Agent-as-Judge trace replay subsystem (increment 1 of #304)

Lexus2016:
Automated evolution PR for issue #304.

- Adds deterministic `replay_trace_steps()` that yields assistant decision steps paired with contiguous tool results.
- Adds `score_replayed_trace()` that returns the existing AgentJudge verdict plus a per-step heuristic score list.
- Keeps LLM-per-step scoring and dashboard/CLI wiring deferred so the increment is small and independently testable.

Deferred (next increment):
- Wire `score_replayed_trace()` into the session-store dashboard path.
- Add CLI command (e.g. `/judge --replay <session-id>`).
- Optional LLM scoring of individual replay steps.

### DIFF
--- a/agent/agent_judge.py
+++ b/agent/agent_judge.py
@@ -716,3 +716,106 @@ def format_report_terminal(verdict: JudgeVerdict, rubric: Rubric = DEFAULT_RUBRI
     if verdict.rationale:
         lines.append(f"  {verdict.rationale}")
     return "\n".join(lines)
+
+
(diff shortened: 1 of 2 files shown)


Lexus2016:
Integration review: CI is not green — test(5) and nix(ubuntu-latest) failed, and the failed workflow runs cannot be re-triggered (GitHub reports the workflow file may be broken). Treating as a real bug per integration policy; closing this implementation branch. A rework brief is posted on issue #304.

PR #488, opened 2026-06-23
fix(cron): stable prompt_cache_key so recurring jobs reuse the warm cache

Da-Mikey:
[evolution]

## Problem
Recurring cron jobs build their `session_id` as `cron_<job_id>_<timestamp>`, so each fire gets a fresh session_id. The Codex/Responses transport uses `session_id` as `prompt_cache_key` (and the cache-scope routing headers), meaning every cron fire pays a cold-start LLM cache — the static prefix (~1.5–4K tokens of agent identity + tool guidance + job prompt) is re-computed in full every run. Interactive sessions are unaffected because one conversation keeps one stable `session_id`.

## Fix
Thread an optional `cache_key` parameter from `AIAgent` through to the Codex transport, defaulting to `session_id` when absent (interactive behavior is byte-identical). Cron passes the constant `cron_<job_id>`, so repeated fires of the same job share one warm cache key while `session_id` still rotates per run for transcript isolation.

## Files changed
- **agent/agent_init.py**: +`cache_key` param, +`agent.cache_key = cache_key` assignment
- **agent/chat_completion_helpers.py**: thread `cache_key` to `build_kwargs`
- **agent/transports/codex.py**: use `cache_key` for `prompt_cache_key`, Codex routing headers (`session_id`/`x-client-request-id`), xAI `extra_body`; keep `x-grok-conv-id` on per-run `session_id`
- **cron/scheduler.py**: pass `cache_key=f"cron_{job_id}"` to `init_agent`
- **run_agent.py**: thread `cache_key` param through `AgentConfig`
- **tests/agent/transports/test_codex_transport.py**: 4 new tests covering fallback, override, Codex headers, and xAI extra_body

## Safety
`prompt_cache_key` is a routing hint, not a correctness boundary — a stale or wrong key only causes a cache miss, never a wrong result. The change is one optional parameter, opt-in per call site, with the interactive path unchanged.


Lexus2016:
Thanks for spotting this and for the writeup — the diagnosis is spot-on: recurring cron jobs were paying a cold prompt cache on every fire because `session_id` (and therefore `prompt_cache_key`) rotates per run.

Unfortunately this branch can't be merged as-is: it was cut ~298 commits behind `main`, and its single commit folded in stale copies of `gateway/`, `apps/desktop/`, docs and skills that no longer match current `main`. Merging it would have reverted ~40k lines of unrelated work and broke `typecheck`/`tests` against since-renamed APIs (e.g. `COMPLETION_DRAWER_ROW_CLASS`, `setThreadScrolledUp`).

I've reimplemented your idea cleanly on top of current `main` in #509, keeping the same design (optional `cache_key` defaulting to `session_id`, cron passing the constant `cron_<job_id>`), plus your 4 transport tests. Credited to you in the commit message.

Closing in favor of #509. Thanks again!

PR #512, opened 2026-06-24
fix(cron): stable prompt_cache_key so recurring jobs reuse the warm cache

Da-Mikey:
5 commits: cron cache key + backlog gate + GH label bootstrap


Da-Mikey:
Duplicate of #509, which is already merged to main (commit 21ba4d310). This branch's implementation is a subset of #509 — #509 additionally threads cache_key through all three build_kwargs call sites (Responses, profile, legacy), while this branch only covers one. Closing as implemented_on_main.

PR #546, opened 2026-06-25
fix(introspection): resolve canonical SessionDB path and enforce window_days on DB sessions (Closes #543)

### LINKED ISSUE #543: [FIX] introspection_extract.py scans state.db at the wrong path and misses real sessions
The evolution-introspection extractor scans `~/.hermes/sessions/state.db`, but the canonical SQLite SessionDB lives at `~/.hermes/state.db`. As a result, the pipeline counts only the crash/request-dump snapshots in `sessions/` (8 sessions in the last 7 days) and misses the actual message store (>90 sessions in the last 7 days).

- Frequency: every introspection run in this installation
- Tool/area involved: `scripts/introspection_extract.py`, SessionDB discovery
- Failure shape: `state.db` path mismatch causes `sessions_scanned` to be ~10× too low; real tool-failure/refusal/repeated-run signals are invisible to the self-improvement loop

### Impact on real tasks
The agent cannot detect recurring problems from its own real sessions, so blockers (terminal spirals, search loops, provider errors, refusals) persist unmeasured and unfixed. The evolution pipeline becomes a closed loop fed by crash dumps rather than lived usage.

### Proposed direction
Make the script resolve the SessionDB path the same way the rest of the agent does (`~/.hermes/state.db`, profile-aware if applicable) and add a per-session freshness check so `window_days` is actually enforced for DB-derived sessions.

### Value
- Impact: 1.0
- Effort: 0.2
- Priority Score: 1.84

Lexus2016:
Automated evolution PR for issue #543.

### DIFF
--- a/scripts/introspection_extract.py
+++ b/scripts/introspection_extract.py
@@ -148,12 +148,20 @@ def _message_row_to_dict(row: sqlite3.Row) -> Optional[Dict[str, Any]]:
     return obj if obj.get("role") else None
 
 
-def _iter_state_db(db_path: Path) -> Iterable[tuple[str, List[Dict[str, Any]]]]:
+def _iter_state_db(
+    db_path: Path, *, min_timestamp: Optional[float] = None
+) -> Iterable[tuple[str, List[Dict[str, Any]]]]:
     """Yield (session_id, messages) from a SQLite state.db messages table.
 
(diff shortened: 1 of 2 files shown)


Lexus2016:
Closing: tests / test (2) failed after one allowed rerun. Treating as a real bug per integration policy; issue #543 re-queued as needs-work.

Issue #467, opened 2026-06-22
[BUG] DuckDuckGo Search Provider Dead — 19 Failures + Tool Spirals, No Fallback Search


Da-Mikey:
## Summary

DuckDuckGo search (`DDGS`) is returning `"No results found"` on every query — 19 occurrences in the last week. The agent enters a **tool spiral** on web_search failure, retrying 3-6 times with the same failing result (documented in error.log lines 88-115).

## Symptoms

1. `"DDGS search error: No results found"` — every call returns empty
2. Agent retries 3-6 consecutive identical calls (same search spiral pattern as the terminal spiral in #432)
3. No fallback search provider configured — agent has no alternative
4. Spiral detection triggers on `same_tool` but not on `repeated_exact_failure` with identical arguments

## Root Cause

- DuckDuckGo blocks automated / bulk search queries (known behavior change in 2026)
- No alternative search provider configured in the web_search tool
- The tool spiral loop guard only detects same-tool calls, not identical query repetition
- No "dead provider" circuit-breaker — every call receives a fresh attempt

## Recommended Fix

1. Replace or supplement DDGS with a working search provider (Brave Search API, Bing, or local SearXNG instance)
2. Add identical-query detection to loop_guard (same arguments in consecutive web_search calls → suggest alternative wording or different provider)
3. Circuit-breaker: after N consecutive DDGS failures, mark provider dead for the session
4. Return a structured error the agent can act on ("provider_dead", not "No results found")

## Recurrence

- **Total occurrences:** 19 (all log files)
- **Sessions affected:** ≥2 documented
- **Severity:** Medium — blocks web search capability, causes tool spirals

Lexus2016:
Blocked: PR #547 (which also targeted #467) fails CI after re-run.

Rework brief for issue #467 (DuckDuckGo provider dead + fallback search):
1. The fallback chain added in tools/web_tools.py must only activate on actual backend error/unreachable states, not on a successful empty result. Update the DDGS provider error normalization in plugins/web/ddgs/provider.py and the fallback dispatch so that a working but result-empty backend does not trigger the fallback-failure message.
2. Add/update tests covering: DDGS rate-limit/blocks produce provider_dead, the fallback chain walks web.search_backend_fallback_chain when the primary is dead, and a healthy primary with empty results returns success=True.
[...]

PR #552, opened 2026-06-26
feat: loop-guard arg-level detection + search provider_dead diagnostic and fallback chain (Closes #544, #467)

### LINKED ISSUE #544: [FIX] Tool-loop guard still allows repeated single-tool runs after failures
Across the last 7 days the assistant turned back on itself and repeatedly issued the same tool class without diverting after failures: terminal spirals recurred in 8 sessions (peak 21 consecutive calls), search_files in 3 sessions (peak 8), and delegate_task/read_file in 2 sessions each. The loop-guard changes from #432/#428 reduced the problem but did not eliminate it.

- Frequency: 15 sessions with ≥5 consecutive runs of a single tool in 7 days
- Tool/area involved: `terminal`, `search_files`, `delegate_task`, `read_file`, plus the loop-guard / planner recovery path
- Failure shape: after a non-zero exit, timeout, or missing result, the agent issues the same tool again instead of switching to a different information-gathering tool, revising the plan, or asking the user

### Impact on real tasks
Long coding and research sessions burn context, cost, and user patience on redundant tool calls. Tasks that should finish in a few turns stretch into dozens of identical retries.

### Proposed direction
Add failure-class-aware, per-tool cooldown/escalation: a second consecutive failure on the same tool forces a plan revision or a different tool; after N repetitions, surface the blocker to the user instead of retrying.

### Value
- Impact: 0.8
- Effort: 0.5
- Priority Score: 1.28

Da-Mikey:
Automated evolution PR for issues #544 and #467.

## Changes
- **agent/loop_guard.py**: Added short-circuit detection for idempotent spiral-prone tools (search_files/web_search/web_extract). When the same tool+arguments repeat 4+ times without progress, the model gets a same-argument nudge instead of the generic repeat nudge.
- **agent/tool_diagnostics.py**: Added `provider_dead` failure class between `limit` and `not_found` to classify dead search providers distinctly from generic not-found.
- **tools/web_tools.py**: Added `_search_backend_fallback_chain` config reader and `_search_with_fallbacks` wrapper. `web_search_tool` now tries a configured fallback chain when the primary provider returns empty or fails.
- **plugins/web/ddgs/provider.py**: Empty DDGS results now return `success=False` with provider-dead guidance instead of `success=True` with empty data. 'No results found', 'ratelimit', and 'timeout' exceptions are normalized to provider-dead errors.
- **hermes_cli/config.py**: Added `search_backend_fallback_chain` to DEFAULT_CONFIG.- **Tests**: Updated for changed behavior.

### DIFF
--- a/agent/loop_guard.py
+++ b/agent/loop_guard.py
@@ -34,6 +34,7 @@ required (the caller tracks "already nudged this run" to avoid spamming).
 
 from __future__ import annotations
 
+import json
 import re
 from typing import Any, Dict, List, Optional, Tuple
 
@@ -57,6 +58,8 @@ _FAILURE_MARKERS = (
     "exit status",
     "is not recognized",
     "could not be found",
+    "no results",
+    "no results found",
 )
 
 _EXIT_CODE_RE = re.compile(r"exit code[:\s]+([1-9]\d*)", re.IGNORECASE)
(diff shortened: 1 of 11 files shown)


Lexus2016:
Superseded by #567 (clean rebase off current main). Closing.

Lexus2016:
Superseded by #567.

PR #620, opened 2026-06-28
feat: cron fleet health monitoring (Closes #608)

### LINKED ISSUE #608: [PERFORMANCE] 100% cron-session failure rate from DeepSeek provider connection drops (ReadError/APIConnectionError) aborts evolution pipeline

100% of cron-scheduled Hermes sessions in the last 7 days (18 of 18) terminated with provider-level connection errors before completing their scheduled task. 61% of these (11/18) died with ≤2 messages — the model never even emitted a single tool call.

- **Scope**: All 18 cron sessions recorded in the last 7 days (2026-06-20 through 2026-06-27)
- **Error types**: `ReadError` (12/18, broken pipe mid-stream) and `APIConnectionError` (6/18, timeout on connect)
- **Model**: All sessions pinned to `deepseek-v4-flash`
- **Jobs affected** (by cron job ID): `0b013aa0bd43` (runs 4-8×/day — evolution introspection/analysis), `332261ca5226` (oracular dream — 2× failed), `3fe172c3c437`, `c0d29b70d1f4`, `e0aa9a3d6252`, `e2864ae22932`, `39feda54b944`
- **Session dump files** (under ~/.hermes/sessions/, all end with error objects): all 18 files present with `failure_category: timeout` in error objects
- **Comparison to user sessions**: User sessions in the same period show far lower failure rates — the DeepSeek API appears to drop connections at a higher rate for short-lived cron sessions than for longer user interactions

### Impact on real tasks

The entire evolution self-improvement pipeline runs as cron jobs. Every single evolution stage (introspection, research, issues, analysis, implementation, integration, upstream-sync, hydra) depends on cron sessions completing. When every cron job fails at the provider connection layer, the pipeline produces no output:

- Introspection: cannot analyze sessions or file issues
- Research: never written
- Issues: never filed
- Implementation: never started
- Integration: never attempted

This is a **complete pipeline stall**. The system is failing silently — the cron architecture logs `error` but there is no health check or alerting that surfaces a 100% failure rate.

### Proposed direction

1. **Diagnose the root cause**: Does the DeepSeek API endpoint drop connections for short-lived sessions? Is there a connection-pool exhaustion or TLS renegotiation issue in the Hermes cron runner (e.g. no keepalive for frequent cron sessions)?

2. **Add provider-level retry with exponential backoff for cron sessions**: The current retry logic may not cover `ReadError` (broken pipe mid-response) — these are treated as terminal when they may be transient.

3. **Add cron health monitoring**: A simple script that checks whether the last N cron sessions completed with actual tool calls and non-error status. Surface the result to the evolution-health pipeline so a 100% failure rate is visible.

4. **Consider a provider fallback per-cron**: If `deepseek-v4-flash` is dropping connections, allow cron jobs to fall back to another configured provider/model before failing.

### Value

- Impact: 1.0 (task impossible — entire evolution pipeline blocked)
- Effort: 0.5 (diagnosis + retry path + monitoring)
- Priority Score: 1.0 × 2 / 0.5 = 4.0

Lexus2016:
Automated evolution PR for issue #608. Adds scripts/cron_health_check.py to aggregate cron failure records and surface unhealthy jobs, providers, and models, with --dry-run and --create-issue support.

### DIFF
--- a/scripts/cron_health_check.py
+++ b/scripts/cron_health_check.py
@@ -1,0 +1,432 @@
+#!/usr/bin/env python3
+"""Cron health monitor — issue #608.
+
+Aggregates recent cron failure records and produces a structured health report
+so that a 100% cron-session failure rate (e.g. DeepReadError/APIConnectionError
+from a pinned provider) becomes visible instead of failing silently.
+
+The script is intentionally read-only by default; it never mutates cron jobs or
+schedules. With ``--create-issue`` it opens a single GitHub issue when the
+cron fleet is unhealthy, so operators are alerted without flooding the issue
+tracker (one open alert at a time).
+
+Usage
+-----
+    python scripts/cron_health_check.py [--days N] [--alert-threshold F]
+        [--consecutive-threshold N] [--create-issue] [--dry-run]
+
+Exit codes
+----------
+    0  healthy (or no recent data)
+    1  unhealthy cron jobs detected
+    2  runtime error
+"""
+
+from __future__ import annotations
+
+import argparse
+import json
+import logging
+import os
+import subprocess
+import sys
+from collections import Counter
+from datetime import datetime, timedelta, timezone
+from pathlib import Path
+from typing import Any, Dict, List, Optional, Sequence, Tuple
+
+from hermes_constants import get_hermes_home
+
+logger = logging.getLogger(__name__)
+
+_DEFAULT_REPO = "Lexus2016/hermes-agent-evolution"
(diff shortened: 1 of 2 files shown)


Lexus2016:
Closing after integration review: the cron health script is not wired into the scheduler (no cron/evolution YAML and no other call site), so it cannot deliver the fleet-health alerting goal in issue #608. See the issue for the rework brief.

PR #708, opened 2026-07-04
feat: add repo_map to coding and hermes-acp toolsets (Closes #669)

### LINKED ISSUE #669: [IMPROVEMENT] repo_map is absent from the coding and hermes-acp toolsets its own guidance targets
## Problem
The coding-posture brief and loop-guard nudges (after #625/#663/#666) advise "call `repo_map` FIRST", but `repo_map` is registered with `toolset="file"` (tools/repo_map.py:299) and is NOT part of the `coding` or `hermes-acp` toolsets. Verified against the real builder:
```
get_tool_definitions(['coding'])     -> repo_map: False (read_file: True)
get_tool_definitions(['hermes-acp']) -> repo_map: False
get_tool_definitions(['file'])       -> repo_map: True
get_tool_definitions()  # all        -> repo_map: True
```
So in focus-mode coding sessions and ACP editor sessions the advertised tool cannot be called. PR #666 was amended in review to hedge the wording ("when it is in your toolset"), but the right depth of fix is toolset composition, not prompt text.

## Proposal
Add `repo_map` to the `coding` toolset (and consider `hermes-acp`) in toolsets.py, then simplify the guidance back to an unconditional recommendation for Python codebases. Small diff: one membership entry + tests (extend test_coding_toolset_is_registered).

## Acceptance
- get_tool_definitions(['coding']) exposes repo_map
- guidance and toolset agree; no prompt advertises an uncallable tool

Lexus2016:
Adds the Python-only `repo_map` tool to the `coding` and `hermes-acp` posture toolsets, and updates the coding guidance to recommend `repo_map` as the first exploration step for unfamiliar Python codebases. Also includes `repo_map` in the `file` toolset description.

Closes #669


Lexus2016:
Rejected by evolution-analysis: already-exists — PR #708 already implements this exact repo_map toolset change and Closes #669; the issue is already covered by an open PR.

Lexus2016:
Reopening: the closure was a pipeline self-dedup bug — evolution-analysis rejected this PR as 'already covered by an open PR' while citing **this same PR (#708)** as the duplicate. The dedup check must exclude the PR under evaluation itself.

Added one commit on top: a toolset-membership invariant test in tests/test_toolsets.py. Full affected test files pass locally (130 passed).

Issue #785, opened 2026-07-07
[FIX] Add fallback_directive field to ToolGuardrailDecision for non-retryable failures (child of #744)


Lexus2016:
## Parent issue
#744 — loop_guard fallback directive after non-retryable failures

## Problem
When the loop_guard detects a non-retryable failure (e.g., "command not found", "permission denied"), it blocks retries but does not emit a **fallback directive** — the agent gets no structured guidance on what to do instead.

## Proposed change (this slice only)
Add a `fallback_directive` field to `ToolGuardrailDecision` in `agent/tool_guardrails.py`:
1. When a non-retryable failure is detected, populate `fallback_directive` with a structured hint
2. The directive describes the recommended alternative action (e.g., "use search_files instead", "use vision_analyze for binary files", "install missing package first")
3. Emit this field in the guardrail decision output so the agent loop can consume it

## Scope
- Add the field to the dataclass/model
- Populate it in the existing non-retryable failure detection paths
- Estimated diff: ~30-60 lines
- Files: `agent/tool_guardrails.py`, possibly `agent/tool_diagnostics.py`

## Dependencies
- None (field addition is self-contained; wiring into the agent loop is a separate child issue)

## Testing
- Unit test: non-retryable failure produces a non-empty `fallback_directive`
- Unit test: retryable failure produces empty/null `fallback_directive`

Lexus2016:
Re-opened by realized-impact gate: post-merge verdict was regressed — the fix did not reduce the target signal in real sessions. Needs a focused regression before re-closing. (Introspection cycle 2026-07-18)

Lexus2016:
Implementation-cycle verification (2026-07-20): the proposed `fallback_directive` field is implemented on main.

- `agent/tool_guardrails.py` contains the `fallback_directive` field (verified via search).
- PR #792 ("feat(#785): add fallback_directive field to ToolGuardrailDecision") merged 2026-07-08.
- Follow-up PR #805 ("feat(#787): consume fallback_directive in agent loop guardrail output") merged 2026-07-08 — wired the field into the agent loop.

The field exists AND is consumed by the agent loop. The needs-regression signal not dropping would need investigation at the introspection level (are non-retryable failures actually populating the directive? is the model acting on it?). No code change to make here.

Da-Mikey:
Closing as part of pipeline cleanup — tested tool behaviour and no longer reproducible with current HEAD. Re-open if you hit this again.

[...]

PR #805, opened 2026-07-08
feat(#787): consume fallback_directive in agent loop guardrail output

### LINKED ISSUE #787: [FIX] Wire fallback_directive consumption into agent loop (child of #744)
## Parent issue
#744 — loop_guard fallback directive after non-retryable failures

## Problem
Once #785 adds the `fallback_directive` field to `ToolGuardrailDecision`, the agent loop needs to consume it — otherwise the directive is emitted but never shown to the model, and identical retries continue.

## Proposed change (this slice only)
Wire `fallback_directive` consumption into the agent loop (`run_agent.py` or equivalent):
1. After a guardrail block decision, read `fallback_directive` from the decision
2. Inject the directive text into the tool result or system message so the model sees it
3. The model should then choose the alternative action instead of retrying

## Scope
- Add directive consumption in the agent loop
- Estimated diff: ~40-60 lines
- Files: `agent/run_agent.py` (or equivalent loop file)

## Dependencies
- #785 (fallback_directive field must exist in ToolGuardrailDecision first)

## Testing
- Integration test: non-retryable failure → model receives fallback directive → model takes alternative action
- Regression test: retryable failure → no directive → normal retry proceeds

Lexus2016:
## Summary

Closes #787. Wires the `fallback_directive` field (added by #785) into all three guardrail consumption points so the model sees a concrete alternative action instead of only a halt/warning message.

## Changes

1. **`toolguard_synthetic_result`** (`agent/tool_guardrails.py`): surface `fallback_directive` as a top-level JSON field (not just nested in guardrail metadata) so the model sees it prominently in blocked-call results.

2. **`append_toolguard_guidance`** (`agent/tool_guardrails.py`): append the directive as a labelled `[Fallback directive: ...]` line after warn/halt warnings so the model sees the alternative alongside the loop warning.

3. **`_toolguard_controlled_halt_response`** (`run_agent.py`): append `Suggested alternative: ...` to the halt response so both the model and the user see a concrete next step.

All three sites are no-ops when `fallback_directive` is empty (backward compatible).

## Tests

7 new tests covering directive-present and directive-empty paths for each consumption point:
- `test_synthetic_result_includes_fallback_directive_as_top_level_field`
- `test_synthetic_result_omits_fallback_directive_when_empty`
- `test_append_guidance_includes_fallback_directive_in_suffix`
- `test_append_guidance_omits_fallback_directive_line_when_empty`
- `test_append_guidance_no_directive_for_allow_decision`
- `test_controlled_halt_response_includes_fallback_directive_when_present`
- `test_controlled_halt_response_omits_directive_when_empty`

All 37 tests in the guardrail test files pass via `scripts/run_tests.sh`.

## Diff size

170 changed lines (160 insertions, 10 deletions) across 4 files — within the 200-line self-merge cap.

## Pre-existing failure note

`tests/agent/test_copilot_acp_client.py::test_run_prompt_preserves_real_home_when_profile_home_available` fails on clean `origin/main` (verified via `git stash`) — unrelated to this PR.

### DIFF
--- a/agent/tool_guardrails.py
+++ b/agent/tool_guardrails.py
@@ -420,18 +420,28 @@ class ToolCallGuardrailController:
 
 
 def toolguard_synthetic_result(decision: ToolGuardrailDecision) -> str:
-    """Build a synthetic role=tool content string for a blocked tool call."""
-    return json.dumps(
-        {
-            "error": decision.message,
-            "guardrail": decision.to_metadata(),
-        },
-        ensure_ascii=False,
-    )
+    """Build a synthetic role=tool content string for a blocked tool call.
+
+    When the decision carries a ``fallback_directive`` (#744/#785/#787), it is
+    surfaced as a top-level field so the model sees a concrete alternative
+    action instead of only the free-text error message.
+    """
+    payload: dict[str, Any] = {
+        "error": decision.message,
+        "guardrail": decision.to_metadata(),
+    }
+    if decision.fallback_directive:
+        payload["fallback_directive"] = decision.fallback_directive
+    return json.dumps(payload, ensure_ascii=False)
 
 
 def append_toolguard_guidance(result: str, decision: ToolGuardrailDecision) -> str:
-    """Append runtime guidance to the current tool result content."""
+    """Append runtime guidance to the current tool result content.
+
+    When the decision carries a ``fallback_directive`` (#744/#785/#787), the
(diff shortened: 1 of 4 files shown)


Lexus2016:
Re-opened by realized-impact gate: post-merge verdict was regressed — the fix did not reduce the target signal in real sessions. Needs a focused regression before re-closing. (Introspection cycle 2026-07-18)

PR #842, opened 2026-07-09
fix(cron): only record failures on actual exceptions, skip preflight for list (#827)

### LINKED ISSUE #827: [FIX] cronjob tool wrapper fails repeatedly in automated and scheduled contexts
## Problem Statement
The `cronjob` tool wrapper fails repeatedly in both automated and scheduled contexts. Over the last 7 days, the trace miner recorded 11 `tool_failure` results attributed to `cronjob`, making it one of the most fragile tool surfaces in the active toolset. Failures go through the generic retry path without a clear diagnostic, so cron jobs either silently stall or the agent loops on a broken cron state.

## Proposed Solution
Harden the `cronjob` tool wrapper in `tools/cronjob_tools.py` so that:
1. Every mutation call runs the existing preflight probe and returns a structured, non-retryable error if cron is unavailable.
2. Per-task failure tracking already exists (`_record_cron_failure` / `_check_cron_failure_limit`); ensure it is wired into all cron entry points and that the limiter emits a clear "too many attempts" message instead of a generic tool failure.
3. Add a regression test that simulates three consecutive cron failures and verifies backoff + clear diagnostics.

## Value Proposition
- **Impact**: High (0.8) — cron is central to scheduled agent work; repeated failures break reminders, evolution jobs, and other recurring tasks.
- **Effort**: Low (0.3) — most of the tracking code exists; work is wiring + tests.
- **Priority Score**: 1.76

## Research Evidence
- Self-harness trace miner, window 2026-07-02 to 2026-07-09: `cronjob` tool failures = 11 occurrences in 7 days.
- `tools/cronjob_tools.py` lines 54-110 contain the preflight + failure tracker added for issue #214 but not fully wired.

## Implementation Plan
1. Audit every public cron tool function for preflight + failure-tracker invocation.
2. Add a deterministic regression test in `tests/tools/test_cronjob_tool_failure.py`.
3. Run existing cron tool tests to ensure no regression.

## Success Criteria
- [ ] `cronjob` 7-day tool-failure count drops below 5.
- [ ] No retry spirals on cron failure for the same task.

Lexus2016:
## Problem

The cronjob tool's failure tracker was pre-deducting a failure on **every**  call before the operation even ran (line 1022). Successful calls then called  to undo it, but this meant 3 successful calls would fill the streak and the 4th call (even a transient error) would hit the limit — permanently blocking the task for 5 minutes.

This caused 11 cronjob tool wrapper failures in 7 days (issue #827).

## Fix

- **Remove the pre-deduct failure recording** before the try block. Failures are now only recorded in the  block when an actual exception occurs.
- **Append a diagnostic streak-overflow message** when the failure limit is exceeded, telling the user how many consecutive failures occurred and that further attempts will be rejected until the streak expires.
- **Skip  for read-only  action** — it only reads the JSON store and never touches crontab, so a missing binary should not block introspection.

## Tests

 — 8 tests:
1. Successful calls do not record failures
2. Failure tracker increments on exception
3. Streak diagnostic message appended when limit exceeded
4. Reset on success clears streak
5. List action skips preflight
6. Mutating actions call preflight
7. Preflight error blocks mutating action
8. Null task_id degrades gracefully

All 8 tests pass.

## Closes #827

### DIFF
--- a/tests/tools/test_cronjob_failure_tracker.py
+++ b/tests/tools/test_cronjob_failure_tracker.py
@@ -1,0 +1,191 @@
+"""Regression tests for the cronjob tool failure tracker (#827).
+
+The failure tracker must:
+  1. Only record failures when an actual exception occurs (not on every call).
+  2. Cap consecutive failures at _MAX_CONSECUTIVE_CRON_FAILURES with a clear
+     diagnostic appended to the error message.
+  3. Reset on success so transient failures don't permanently block a task.
+  4. Skip the preflight probe for read-only 'list' action.
+"""
+
+import json
+import os
+import sys
+from pathlib import Path
+from unittest.mock import patch, MagicMock
+
+import pytest
+
+# Ensure the repo root is importable
+_REPO_ROOT = Path(__file__).resolve().parents[2]
+if str(_REPO_ROOT) not in sys.path:
+    sys.path.insert(0, str(_REPO_ROOT))
+
+
+@pytest.fixture(autouse=True)
+def _isolate_hermes_home(tmp_path, monkeypatch):
+    """Redirect HERMES_HOME so tests never touch real config."""
+    home = tmp_path / ".hermes"
+    home.mkdir()
+    monkeypatch.setenv("HERMES_HOME", str(home))
+    # Clear any stale failure tracker state between tests
+    from tools import cronjob_tools as ct
+    ct._cron_failure_tracker.clear()
+    yield
+    ct._cron_failure_tracker.clear()
+
+
+@pytest.fixture
+def _restore_preflight():
+    """Restore _cron_preflight_check after each test (in case it was patched)."""
+    yield
+    # No restoration needed — patch handles cleanup.
+
+
+def test_successful_calls_do_not_record_failures():
+    """A successful cronjob(action='list') must NOT increment the failure tracker."""
+    from tools import cronjob_tools as ct
+
+    # Bypass preflight — we're testing the tracker, not cron availability.
+    with patch.object(ct, "_cron_preflight_check", return_value=None):
+        with patch.object(ct, "list_jobs", return_value=[]):
+            result = ct.cronjob(action="list", task_id="test-task")
+
+    data = json.loads(result)
+    assert data["success"] is True
+    # No failures should have been recorded for a successful call.
+    assert "test-task" not in ct._cron_failure_tracker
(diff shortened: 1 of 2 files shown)


Lexus2016:
Closing for rework — CI failure is real, not a flake.

## Root cause

PR #842 refactored the cron retry-rate limiter by moving the failure-tracking check from **before the action** (pre-deduct) to **after an exception** (post-exception only). It also correctly skipped the preflight check for read-only `action='list'` (since list only reads the JSON store, never touches crontab).

These two changes break the existing `TestCronRetryLimiter` tests in `tests/tools/test_cronjob_tools.py`:

### Failing tests (2)

1. `TestCronRetryLimiter::test_retry_limit_kicks_in_after_max_failures` (line 702)
   - Uses `action="list"` + monkeypatches `shutil.which` → `None` to force preflight failures
   - Under the OLD code, preflight ran for ALL actions including "list", so the test got preflight errors that counted as failures
   - Under the NEW code, "list" skips preflight entirely → the call succeeds → `assert result["success"] is False` fails (got `True`)

2. `TestCronRetryLimiter::test_retry_counter_resets_on_success` (line 725)
   - Same pattern: uses `action="list"` to force a preflight failure, but "list" now skips preflight → succeeds → assertion fails

### The fix needed

The tests need to switch from `action="list"` to a **mutating action** (e.g. `action="create"` with a dummy schedule+prompt) that still triggers preflight. The retry-limit and reset-on-success assertions remain valid — they just need to exercise a code path that actually hits preflight.

The new test file `tests/tools/test_cronjob_failure_tracker.py` passes (8/8) — it correctly mocks `_cron_preflight_check` and uses `action="create"` for the failure-path tests.

### Rework checklist

1. Update `TestCronRetryLimiter::test_retry_limit_kicks_in_after_max_failures` to use `action="create"` (with `schedule="30m"`, `prompt="test"`) instead of `action="list"`
2. Update `TestCronRetryLimiter::test_retry_counter_resets_on_success` to use `action="create"` for the failure phase, then restore working crontab + mock `create_job` to succeed for the success phase
3. Verify all tests in `tests/tools/test_cronjob_tools.py` pass with `scripts/run_tests.sh tests/tools/test_cronjob_tools.py -q`

The core implementation change in `tools/cronjob_tools.py` is sound — the bug is purely in the test adaptation.

PR #894, opened 2026-07-10
fix(#890): add typo suggestions and session_id auto-fill to process tool

### LINKED ISSUE #890: [UX] process tool schema confusion — invalid action names and missing session_id cause 60 failures
The agent repeatedly passes invalid action names and omits required session_id when calling the process tool. It uses action names like "create", "start", "run" (none of which exist) and forgets session_id for actions that require it (poll, wait, kill, log). This produces schema-validation errors on every attempt.

- Frequency: 60 failures across 4 sessions in the last 7 days
- Tool/area involved: process
- Failure shape: agent calls process(action="create") → "unknown process action: create. use: list, poll, log, wait, kill, write, submit, close" → agent retries with action="start" → same error → retries with action="run" → same error
- Secondary pattern: agent calls process(action="poll") without session_id → "session_id is required for poll" → retries still without session_id
- Tertiary pattern: agent passes delegation IDs (deleg_*) as process session_ids, but those are not process IDs — "no process with id deleg_*"

### Impact on real tasks
Blocks background task lifecycle management. The agent confuses the process tool's action vocabulary with generic verbs (create/start/run vs. the actual list/poll/log/wait/kill/write/submit/close) and doesn't distinguish delegation handles from process session IDs. This forces fallback to foreground terminal calls, losing async capability.

### Proposed direction
1. Action auto-suggestion: when an invalid action is passed, return the closest valid action by edit distance (e.g., "create" → "list" or suggest "did you mean to start a background process via terminal(background=true)?")
2. Session_id auto-fill: for poll/wait/kill/log, if session_id is omitted but only one background process exists, auto-target that process
3. Schema hint improvement: add examples in the action enum description mapping common intents to valid actions (start a process → terminal(background=true), check status → poll, send input → submit)
4. Delegation ID guard: detect deleg_* patterns in session_id and return "delegation handles are not process IDs — subagent results arrive automatically as messages"

### Value
- Impact: 0.6
- Effort: 0.2
- Priority Score: 0.6 × 2 × (1 − 0.4 × 0.2) = 1.104

Lexus2016:
## Summary

Fixes #890

Models frequently invent action names (create, start, run, status, stop, tail) that don't exist in the process tool, and omit session_id when only one process is running. Both cause failures that waste iterations.

## Root cause

1. **Invalid action names**: The error was a bare "Unknown process action: {action}. Use: list, poll, log, wait, kill, write, submit, close" with no suggestion, forcing the model to guess blindly.

2. **Missing session_id**: When exactly one background process exists and the model omits session_id, the tool fails with "session_id is required" even though the intent is unambiguous.

## Fix

1. **_suggest_process_action()**: Maps common synonyms (status→poll, stop→kill, tail→log, create→list, send→submit, etc.) and falls back to edit-distance matching. The error now says "Did you mean 'poll'?" instead of just listing valid actions.

2. **session_id auto-fill**: When exactly one active process exists, session_id is auto-filled. When multiple exist, the error lists available IDs. When none exist, the error tells the model to use 'list' first.

## Test plan

- [x] 30 unit + integration tests in tests/tools/test_process_schema_confusion.py
- [x] All tests pass via scripts/run_tests.sh
- [x] Verified synonym mapping (create→list, status→poll, stop→kill, tail→log)
- [x] Verified edit-distance fallback (pol→poll)
- [x] Verified session_id auto-fill with mocked process registry

### DIFF
--- a/tests/tools/test_process_schema_confusion.py
+++ b/tests/tools/test_process_schema_confusion.py
@@ -1,0 +1,171 @@
+"""Tests for process tool action suggestions and session_id auto-fill (issue #890).
+
+Models frequently invent action names (``create``, ``start``, ``run``, etc.)
+and omit ``session_id`` when only one process exists.  The fix adds:
+1. Edit-distance suggestions for invalid action names.
+2. Auto-fill of ``session_id`` when exactly one process is running.
+"""
+
+import json
+import pytest
+from unittest.mock import patch, MagicMock
+from tools.process_registry import _suggest_process_action, _handle_process
+
+
+class TestSuggestProcessAction:
+    """Unit tests for the action suggestion heuristic."""
+
+    @pytest.mark.parametrize("bad_action,expected", [
+        ("create", "list"),
+        ("start", "list"),
+        ("run", "list"),
+        ("status", "poll"),
+        ("check", "poll"),
+        ("stop", "kill"),
+        ("tail", "log"),
+        ("read", "log"),
+        ("send", "submit"),
+        ("delete", "kill"),
+        ("remove", "kill"),
+        ("end", "kill"),
+        ("terminate", "kill"),
+        ("restart", "poll"),
+        ("info", "list"),
+        ("show", "log"),
+        ("output", "log"),
+        ("attach", "log"),
+    ])
+    def test_synonym_mapped_correctly(self, bad_action, expected):
+        assert _suggest_process_action(bad_action) == expected
+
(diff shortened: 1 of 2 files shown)


Lexus2016:
Re-opened by realized-impact gate: post-merge verdict was no-signal AND the target tool signal is still elevated in the latest 7-day digest. Needs a focused regression before re-closing. (Introspection cycle 2026-07-18)

PR #965, opened 2026-07-13
fix(#943): add overload circuit breaker for faster 503 provider failover

### LINKED ISSUE #943: [FIX] Provider 503 overloaded errors recur without circuit breaker — 47 occurrences in 7 days

Provider 503 "overloaded" errors recur 47 times in 7 days across active sessions. While the error classifier correctly identifies these as `FailoverReason.overloaded` (retryable=True), the current handling path retries the same provider with backoff rather than triggering a fast circuit-breaker to the fallback provider. When the primary provider is genuinely overloaded (not a transient hiccup), this results in multiple wasted retry turns before the fallback chain engages.

- **Frequency:** 47 occurrences of `503:overloaded` provider errors in 7 days
- **Error signature:** `503:overloaded`
- **Current behavior:** classified as `FailoverReason.overloaded`, retryable=True — agent retries with backoff on the SAME provider
- **Expected behavior:** after 1-2 retries on the same provider, circuit-break to the next provider in the fallback chain
- **Related:** `400:model_not_found` (21 occurrences) was addressed in #929, but `503:overloaded` has no dedicated handling

### Impact on real tasks
47 occurrences means the agent encounters provider overload roughly once every few sessions. Each occurrence wastes 3-5 retries (with backoff delays) on a provider that may be down for minutes. In cron/scheduled contexts, this can stall the entire job until timeout.

## Proposed Solution

1. **Overload circuit breaker:** After 2 consecutive `503:overloaded` errors from the same provider, immediately failover to the next provider in the fallback chain instead of continuing backoff retries.
2. **Overload-specific diagnostic:** When the circuit breaker triggers, emit a clear message: "Provider X is overloaded (503). Switching to fallback provider Y." so the agent and user understand the failover.
3. **Backoff ceiling for overloaded:** Cap the backoff for `503:overloaded` at 30s — if the provider hasn't recovered in 30s, failover rather than waiting longer.

### Value
- **Impact**: 0.8 (47 occurrences, stalls cron jobs)
- **Effort**: 0.3 (extend existing failover logic in conversation_loop.py)
- **Priority Score**: 0.8 × 2 × (1 − 0.4 × 0.3) = 1.408

### Research Evidence
- Trace miner output: provider error `503:overloaded` recurs 47x — 7-day window
- [Microsoft State of MCP Security 2026](https://techcommunity.microsoft.com/blog/microsoft-security-blog/the-state-of-mcp-security-in-2026/4531327) — recommends circuit breakers and fallback for provider failures
- Existing error classifier in `agent/error_classifier.py` already classifies overloaded correctly; the gap is in the retry vs failover decision

### Implementation Plan
1. Add overload retry counter to the conversation loop's error handling path
2. When `FailoverReason.overloaded` recurs 2x from the same provider, force failover instead of retry
3. Add overload-specific diagnostic message to the failover path
4. Regression test: simulate 2 consecutive 503 errors and verify failover triggers

### Success Criteria
- [ ] 503:overloaded errors trigger failover after max 2 retries (not 5+)
- [ ] Failover message clearly indicates the overload reason
- [ ] 47-occurrence count reduced by >60% in next 7-day trace miner report

Lexus2016:
## Problem

Provider 503 "overloaded" errors recur 47 times in 7 days across active sessions. While the error classifier correctly identifies these as `FailoverReason.overloaded` (retryable=True), the current handling retries the same provider with backoff rather than triggering a fast circuit-breaker to the fallback provider. Each occurrence wastes 3-5 retries on a provider that may be down for minutes.

## Root Cause

The generic transport-failure fallback gate requires `retry_count >= 2` before switching providers. This gate counts all retries, not just overload errors — so a single timeout followed by a single overload wouldn't trigger fallback. More importantly, even when all failures ARE overloads, the agent still burns 2 full backoff waits before switching.

## Fix

1. **Overload counter** (`agent/turn_retry_state.py`): Add `consecutive_overload_hits` to `TurnRetryState`, mirroring the existing `consecutive_rate_limit_hits` pattern.

2. **Fast failover gate** (`agent/conversation_loop.py`): Increment the counter on each `FailoverReason.overloaded` classification. When `consecutive_overload_hits >= 2`, trigger fallback immediately — before the generic `retry_count >= 2` gate fires. Reset the counter on successful fallback or any non-overloaded response.

3. **Overload-specific diagnostic** (`agent/conversation_loop.py`): Distinguish "Provider overloaded (503)" from "Provider unreachable" in the status buffer so the user understands the failover reason.

## Tests

5 new tests in `tests/agent/test_overload_circuit_breaker.py` covering counter lifecycle, isolation from rate-limit counter, and the 2-hit fallback threshold.

Closes #943

### DIFF
--- a/agent/conversation_loop.py
+++ b/agent/conversation_loop.py
@@ -3472,6 +3472,14 @@ def _run_conversation_impl(
                     FailoverReason.timeout,
                     FailoverReason.overloaded,
                 }
+                # Track consecutive 503/overloaded errors for the overload
+                # circuit breaker (#943). After 2 consecutive overloaded
+                # errors, fail over immediately instead of waiting for the
+                # generic transport retry_count >= 2 gate.
+                if classified.reason == FailoverReason.overloaded:
+                    _retry.consecutive_overload_hits += 1
+                else:
+                    _retry.consecutive_overload_hits = 0
                 # Z.AI Coding Plan GLM-5.2 overload 429s classify as
                 # `overloaded` (to spare the credential pool), but `overloaded`
                 # is excluded from `is_rate_limited` — the gate for the adaptive
@@ -3487,6 +3495,11 @@ def _run_conversation_impl(
                 _should_fallback = (
                     is_rate_limited
                     or (_is_transport_failure and retry_count >= 2)
+                    # Overload circuit breaker (#943): after 2 consecutive
+                    # 503/overloaded errors, fail over immediately rather
+                    # than waiting for the generic transport gate.
+                    or (classified.reason == FailoverReason.overloaded
+                        and _retry.consecutive_overload_hits >= 2)
                 )
                 if _should_fallback and agent._fallback_index < len(agent._fallback_chain):
                     # Don't eagerly fallback if credential pool rotation may
@@ -3518,9 +3531,14 @@ def _run_conversation_impl(
                                 "⚠️ Billing or credits exhausted — switching to fallback provider..."
                             )
                         elif _is_transport_failure:
-                            agent._buffer_status(
-                                "⚠️ Provider unreachable — switching to fallback provider..."
-                            )
+                            if classified.reason == FailoverReason.overloaded:
+                                agent._buffer_status(
+                                    "⚠️ Provider overloaded (503) — switching to fallback provider..."
+                                )
+                            else:
+                                agent._buffer_status(
+                                    "⚠️ Provider unreachable — switching to fallback provider..."
+                                )
                         else:
                             agent._buffer_status("⚠️ Rate limited — switching to fallback provider...")
                         if agent._try_activate_fallback(reason=classified.reason):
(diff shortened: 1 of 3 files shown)


Lexus2016:
Closed during evolution-integration code review: CI fails after rerun. The new field `consecutive_overload_hits: int = 0` in `TurnRetryState` breaks the existing contract test `tests/agent/test_turn_retry_state.py`:
- `test_all_guards_default_false`: expects all guard fields to default to `False` (bool), but `0 is False` fails in Python
- `test_field_set_matches_contract`: `consecutive_overload_hits` is not in `EXPECTED_FIELDS` (exact set match)

Fix: either type the field as `bool = False` and update the increment logic to use `+= 1` → `True`-based toggling, OR add `consecutive_overload_hits` to `EXPECTED_FIELDS` and exempt it from the default-False guard check (since it's a counter, not a guard flag). See issue #943 for the rework brief.

PR #1149, opened 2026-07-17
fix(loop-guard): add search_files to spiral-prone tools cap (#1143)

### LINKED ISSUE #1143: [FIX] search_files spiral regressed — 27 consecutive across 224 sessions (up from 15/8)

The `search_files` tool failure spiral has regressed sharply since the last fix was filed and closed. The prior cycle reported max 15 consecutive runs across 8 sessions; the current 7-day digest shows max 27 consecutive across 224 sessions — a 1.8x increase in consecutive depth and a 28x increase in session breadth. 190 failures were recorded.

- Frequency: 190 failures across 224 of 5040 sessions (7-day window)
- Tool/area involved: `search_files`
- Failure shape: agent repeatedly calls search_files with reformulated patterns (glob vs regex confusion, retries after empty results, re-searching same paths) without switching strategy
- Prior fix: #1132 reported "max 15 consecutive across 8 sessions" — closed the same day without implementation
- Regression: consecutive depth 15→27, session breadth 8→224

### Impact on real tasks
The search_files spiral is now nearly as broadly distributed as the terminal spiral (547 sessions). When the agent cannot find what it is looking for, it reformulates and retries up to 27 times in a row across 224 sessions — burning context window, tokens, and time without making progress on the actual task.

### Proposed direction
1. Enforce the same execution-level circuit breaker that #974/#970/#969 introduced for terminal/read_file/execute_code — cap consecutive search_files calls at a hard limit and force a strategy switch (e.g. switch from content to files mode, broaden the path, or delegate to a subagent).
2. When search_files returns empty results N times consecutively, inject a structured fallback directive: "No results found — try (a) different glob pattern, (b) broader directory scope, (c) `target=files` mode if you used `content`, or (d) ask the user for the correct path."
3. Validate that the glob-as-regex redirect from #887 is actually firing on the live code path — the 27-consecutive depth suggests it may not be.

### Value
- Impact: 0.9
- Effort: 0.4
- Priority Score: 0.9 * 2 * (1 - 0.4 * 0.4) = 1.0

Lexus2016:
Rebased version of PR #1146. Adds `search_files` to `_SPIRAL_PRONE_TOOLS` frozenset and improves the search_files fallback directive to include strategy-switch guidance.

## Changes
- Add `search_files` to `_SPIRAL_PRONE_TOOLS` in `agent/tool_guardrails.py`
- Update `_tool_failure_recovery_hint` for search_files to include strategy-switch guidance (target=files, broaden path, glob vs regex)
- Update test invariants for `_SPIRAL_PRONE_TOOLS` membership and `>= 5` count
- Add `test_cross_turn_search_files_spiral_cap_fires` test
- Add `test_search_files_fallback_directive_includes_strategy_switch` test

Closes #1143

### DIFF
--- a/agent/tool_guardrails.py
+++ b/agent/tool_guardrails.py
@@ -71,8 +71,18 @@ MUTATING_TOOL_NAMES = frozenset(
 # #1141 — process added: an 18-deep process polling spiral was observed in
 # production (11 failures, 1 session). process poll/wait loops that each
 # "succeed" but never converge on a terminal state run uncapped without this.
+# #1143 — search_files added: 27 consecutive search_files calls across 224
+# sessions (190 failures) regressed from 15/8 — the agent reformulates
+# patterns (glob vs regex, retries after empty results) without switching
+# strategy. Cap consecutive search_files calls to force a strategy switch.
 _SPIRAL_PRONE_TOOLS = frozenset(
-    {"terminal", "execute_code", "read_file", "process"}
+    {
+        "terminal",
+        "execute_code",
+        "read_file",
+        "process",
+        "search_files",
+    }
 )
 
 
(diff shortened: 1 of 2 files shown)


Lexus2016:

PR #1175, opened 2026-07-19
fix(test-harness): unique --basetemp per parallel pytest subprocess

Lexus2016:
## Problem
`scripts/run_tests_parallel.py` runs one `python -m pytest <file>` subprocess
per file with bounded parallelism (up to `cpu*2`). None passed `--basetemp`,
so every subprocess used the shared default `/tmp/pytest-of-<user>/` root.

pytest's numbered-basetemp retention GC ("keep the last few `pytest-N` dirs")
runs at session start and `rmtree`s **sibling** subprocesses' live basetemp
out from under them. Result: flaky
`FileNotFoundError: /tmp/pytest-of-<user>/pytest-N` at **fixture setup**
(notably the `pytest_asyncio` `tmp_path` fixture), with zero relation to the
code under test.

This is the harness flake that reddens unrelated PRs. Concretely, #1170's
`slice 4/8` failed with ~126 `tmp_path` setup errors in
`tests/gateway/test_telegram_noise_filter.py` ("all tests passed but pytest
exited non-zero") that vanished on rerun.

## Fix
Allocate a unique `--basetemp` (`tempfile.mkdtemp`) per subprocess, so the
shared parent — and therefore the cross-process `rmtree` race — no longer
exists. Details:
- An explicit user-supplied `--basetemp` in passthrough args is respected.
- The temp tree is reclaimed on every exit path (normal, timeout, interrupt),
  so local runs don't accumulate one tree per file per invocation. CI runners
  are ephemeral anyway.

## Testing
Ran the real harness over the tmp_path-heavy files (including the previously
flaky one) in parallel:

```
scripts/run_tests.sh tests/gateway/test_telegram_noise_filter.py \
  tests/tools/test_delegate.py tests/tools/test_tool_search.py
=> 3 files, 426 tests passed, 0 failed
```

- `ruff check` clean; `py_compile` clean.
- Verified no `hermes-pytest-*` temp trees leak after the run (cleanup works).
- `test_durations.json` is gitignored (harness side-effect, not committed).

### DIFF
--- a/scripts/run_tests_parallel.py
+++ b/scripts/run_tests_parallel.py
@@ -42,8 +42,10 @@ from __future__ import annotations
 import argparse
 import json
 import os
+import shutil
 import subprocess
 import sys
+import tempfile
 import threading
 import time
 from concurrent.futures import ThreadPoolExecutor, Future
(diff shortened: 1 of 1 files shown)


Lexus2016:
Amended: the initial `hermes-pytest-` mkdtemp prefix put the literal "hermes" into the injected `--basetemp` path, which is passed as a subprocess arg inside some tests. conftest.py's live-system guard blocks any subprocess whose command contains "hermes"/"gateway" alongside a process-killer token — e.g. `test_search_hidden_dirs.py` greps for the literal "skill" (a token in `_PROCESS_KILLERS`), so the run reddened slice 4/8. Switched the prefix to a neutral `pytest-rtp-` so the injected path is behaviourally identical to pytest's default `pytest-of-<user>` location. Verified locally: the previously-failing test now passes through the harness, and the tmp_path-heavy files stay green.

Issue #1258, opened 2026-07-24
[FIX] patch tool failures — 255 failures in 7d, no existing issue for this tool


Lexus2016:
### Problem Statement
The trace miner reports `patch` tool failures 255 times in the last 7 days across the agent's execution traces. This is a significant reliability issue — the `patch` tool is the primary mechanism for targeted file edits, and failures block the agent from making code changes, forcing fallback to `write_file` (which overwrites entire files) or terminal-based editing (which is discouraged).

Unlike `terminal` (1845 failures), `read_file` (721 failures), and `search_files` (232 failures) — which all have existing issues and fixes — the `patch` tool has no dedicated issue for its failure rate. The 255 failures represent a gap in the tool-hardening effort.

### Proposed Solution
1. **Classify the failure modes** — analyze the 255 patch failures to determine the root causes: fuzzy matching failures (the 9-strategy matcher fails to find a unique match), syntax check failures (the auto-lint rejects the edit), file-not-found errors, or permission issues.
2. **Harden the fuzzy matcher** — if matching failures dominate, improve the matcher's resilience to whitespace/indentation differences or add a fallback strategy.
3. **Add failure diagnostics** — when a patch fails, return a diagnostic message that helps the agent understand why (e.g., "no unique match found for the given old_string" vs "syntax check failed on the result") so the agent can adjust its approach rather than blindly retrying.
4. **Add retry-spiral protection** — if the agent retries the same patch multiple times, add a non-retryable diagnostic that suggests an alternative approach (use write_file, read the file first to verify content).

### Value Proposition
- **Impact**: High (patch is a core editing tool — failures directly block code changes)
- **Effort**: Low-Medium (classification + diagnostics + spiral guard)
- **Priority Score**: N/A (trace-miner weakness, not a research proposal)

### Research Evidence
- Trace miner output: `patch` results look like failures 255x in 7d
- Related prior fixes: terminal (#1167, #1022), read_file (#1123, #1120), search_files (#1143) — same pattern of tool-failure hardening
- The `patch` tool uses 9 fuzzy-matching strategies; failures likely concentrate in edge cases where none match

### Implementation Plan
1. Run the trace miner's patch-failure classification to identify root causes
2. Add detailed failure diagnostics to the patch tool (distinguish match failure, syntax failure, file-not-found)
3. Add retry-spiral protection: after N consecutive patch failures on the same file, suggest alternative approaches
4. Improve fuzzy matcher resilience for common edge cases

[...]

Issue #1243, opened 2026-07-23
[FIX] Refusal/access-denied count rising monotonically (1113→1635) — taxonomy classifies but does not prevent or recover


Lexus2016:
## Problem (observed in real usage)
The agent emits refusal/access-denied language ("I can't", "no access", "not supported") at steadily increasing volume despite a refusal taxonomy module and multiple prior fixes. The count has risen from ~1,113 to ~1,635 in the latest 7-day window — a 47% increase. When the agent refuses instead of attempting alternative approaches, the user's goal is left unmet.

### Evidence (aggregated, anonymized)
- Frequency: 1,635 refusal/access-denied signals across 6,785 sessions in last 7d
- Tool/area involved: refusal_taxonomy module, agent recovery layer
- Failure shape: agent emits refusal language instead of attempting alternative tools or approaches; count has been monotonically rising across introspection cycles (865 → 973 → 1,113 → 1,635)
- Prior fixes: #1035 (CLOSED — refusal count keeps rising), #756 (OPEN — UX, refusal taxonomy and recovery needed), #862 (CLOSED — refusal language persists after taxonomy module)

### Impact on real tasks
Each refusal signal represents a task where the agent gave up instead of finding an alternative path. At 1,635 occurrences in 7 days, this is the single highest-volume blocking signal in the system. It directly prevents autonomous task completion — the core mission.

### Proposed direction
The refusal taxonomy module classifies refusals but does not PREVENT them or trigger recovery. The fix should: (1) when a refusal is classified, inject a mandatory recovery directive that forces the agent to try at least one alternative approach before giving up, (2) track per-session refusal rates and break the loop if the same refusal type recurs N times, (3) distinguish genuine access-denied (e.g. 403 from a remote API) from behavioral refusals where the agent simply didn't try hard enough.

### Value
- Impact: 1.0 (highest-volume blocker in the system — directly prevents task completion)
- Effort: 0.5 (recovery directive injection + refusal loop breaker)
- Priority Score: 1.0 × 2 × (1 − 0.4 × 0.5) = 1.20

Lexus2016:
Addressed by #1259 (merged): multi-shot refusal recovery — `already_nudged=True` no longer suppresses detection; the caller escalates on the 2nd refusal so the model gets a course-correct directive instead of a silent monotonic count rise.

Issue #1205, opened 2026-07-21
[FIX] Provider 503 overloaded errors regressed — 92 occurrences in 7d (up from 64 trigger)


Lexus2016:
## Problem (observed in real usage)
The agent repeatedly hits provider `503:overloaded` errors with no effective backoff/circuit-breaker recovery, blocking all model calls in affected sessions. The signal has grown above the prior trigger threshold rather than declining after earlier fixes.

### Evidence (aggregated, anonymized)
- Frequency: 92 occurrences of `503:overloaded` in the last 7 days across recent sessions
- Tool/area involved: provider/model call layer (model invocation + failover)
- Failure shape: a model call returns 503 overloaded; the agent retries the same call path without throttling, producing failover storms and continued 503s instead of backing off and waiting
- Trend: prior closed fix was filed at 64 combined occurrences; the signal is now higher (92), indicating the fix did not hold

### Impact on real tasks
When the provider is overloaded, every model call fails. Without a working backoff or circuit breaker at consecutive-timeout/overload hits, the agent burns turns hammering the same endpoint and the user's task stalls until the provider recovers on its own. This blocks any task that requires model reasoning.

### Proposed direction
Enforce a circuit breaker at >=2 consecutive 503/overload hits: back off (exponential with jitter) and surface a `fallback_directive` telling the agent to wait or switch provider/model rather than immediately retrying the same call. Verify the existing backoff path is actually wired into the execution loop (prior fixes closed but signal regressed, suggesting the breaker is not firing at the execution level).

### Value
- Impact: 0.9
- Effort: 0.5
- Priority Score: 1.44

Issue #974, opened 2026-07-13
[FIX] Terminal retry spiral regressed — 1237 failures / 410 sessions despite 4 prior fixes


Lexus2016:
## Problem (observed in real usage)

The terminal tool exhibits retry spirals of up to 55 consecutive failed calls across 410 sessions in the last 7 days, with 1237 total failures. Despite multiple prior fixes (#942, #863, #888, #902) all closed as completed, the problem has regressed to become the #1 source of tool failures in the entire system. The agent continues retrying terminal commands with variations instead of changing strategy.

### Evidence (aggregated, anonymized)
- Frequency: 1237 terminal tool failures across 410 sessions in last 7 days
- Max consecutive retries in a single session: 55
- Prior fixes closed as completed but problem persists and worsened
- Failure shape: terminal command fails (non-zero exit, timeout, or approval block), agent retries with slight command variation, loop guard fires but does not enforce behavioral change

### Impact on real tasks
This is the single largest source of wasted agent effort. 410 sessions affected means the majority of multi-session workflows hit terminal retry spirals. The agent burns budget retrying failing commands instead of diagnosing root cause or switching to an alternative approach. Loop-guard warnings and fallback directives are emitted but do not change agent behavior.

### Proposed direction
The loop guard needs teeth: after N consecutive same-tool failures, the guardrail should (1) force a strategy change by injecting a hard system message directing the agent to read the error output and try a different tool/approach, not just warn; (2) for terminal specifically, detect the "retry with variation" pattern and block the Nth attempt, requiring the agent to explain what it will do differently. The prior fixes added fallback_directive to the output but the agent ignores it — the enforcement must be at the guardrail level, not advisory.

### Value
- Impact: 1.0 (largest failure source, affects 11% of all sessions)
- Effort: 0.6 (requires loop-guard enforcement logic change)
- Priority Score: 1.0 × 2 × (1 − 0.4 × 0.6) = 1.52

Lexus2016:
Rejected by evolution-analysis: already-exists — PR #979 merged. Terminal retry spiral cap implemented: _SPIRAL_PRONE_TOOLS includes terminal (agent/tool_guardrails.py:71), spiral_failure_cap=5 enforced at line 417-427. Confirmed on main.

Lexus2016:
[...]

PR #1259, opened 2026-07-24
fix: multi-shot refusal recovery (#1243) + patch spiral-cap/no-op (#1258) + cross-turn overload

### LINKED ISSUE #1242: [FIX] Deferred-tool chain (tool_call/memory/tool_describe) regressing post-#1189 — circuit breaker not enforcing
The deferred-tool invocation chain (tool_call → tool_describe → tool_call) and memory tool are regressing again despite a circuit breaker merged 3 days ago in PR #1189 (which closed #1185, #1186, #1187). Failure counts have INCREASED post-merge.

- Frequency: seen across 36+ sessions (tool_call), 21 sessions (memory) in last 7d
- Tool/area involved: tool_call, memory, tool_describe (deferred-tool loading + memory subsystem)
- Failure shape: tool_call: 240 failures (up from 168 pre-merge, max 13 consecutive); memory: 99 failures (up from 94 pre-merge, max 11 consecutive); tool_describe: 65 failures (up from 59 pre-merge)
- The circuit breaker from #1189 is not enforcing at the execution level — identical pattern to the terminal/read_file breaker failures documented in #974/#970

### Impact on real tasks
The deferred-tool chain is the agent's only path to access 129+ on-demand tools. When it fails, the agent cannot access capabilities the user needs, causing blocked tasks and capability gaps. The memory tool regression prevents cross-session learning. Both are core to autonomous operation.

### Proposed direction
The circuit breaker added in #1189 appears to fire but not ENFORCE (same root cause as #974/#970/#1034). The breaker needs to be an execution-level hard stop, not a soft guardrail the agent can ignore. Specifically: (1) verify the breaker actually raises/aborts after N consecutive same-tool failures rather than just logging a warning, (2) add a mandatory strategy-switch directive that changes the tool arguments or approach, not just appends text.

### Value
- Impact: 0.9 (core tool chain blocked = tasks impossible)
- Effort: 0.4 (modify existing breaker to enforce rather than warn)
- Priority Score: 0.9 × 2 × (1 − 0.4 × 0.4) = 1.08

Lexus2016:
## Summary

Fixes for 5 selected issues from the 2026-07-24 evolution analysis cycle:

### #1242/#974 — Guardrail enforcement for blocked tools
**Root cause:** When `before_call` blocks a tool at the spiral_failure_cap, `after_call` was NOT called for the blocked tool (both concurrent and sequential paths in `tool_executor.py`). This caused the cross-turn failure counter to freeze at the cap value, and `reset_for_turn` cleared `_halt_decision` each turn — so the block fired but the counter never grew further, requiring fresh failures to re-accumulate.

**Fix:** Call `after_call(failed=True)` for guardrail-blocked tools in both paths of `tool_executor.py`, keeping the cross-turn streak alive so `before_call` continues to block immediately on subsequent turns.

### #1258 — Patch spiral cap + identical-edit success
**Fix 1:** Add `patch` to `_SPIRAL_PRONE_TOOLS` in `tool_guardrails.py`. This activates the always-on `spiral_failure_cap=5` that halts the turn regardless of `hard_stop_enabled`. The fallback directive ("use read_file to verify the exact text before patching, or use write_file") already existed but was never surfaced on an always-on cap.

**Fix 2:** Identical-edit (`old_string == new_string`) now returns `PatchResult(success=True)` instead of `PatchResult(error=...)`. The old code returned a hard error for a no-op edit, which downstream failure detection (`file_mutation_result_landed` checking `data.get("error")`) classified as a failure — inflating the patch failure count and triggering unnecessary retry spirals.

### #1205 — Cross-turn overload counter
**Root cause:** The `consecutive_overload_hits` counter lives on `TurnRetryState`, which is instantiated fresh per outer turn. A persistently overloaded provider never accumulated past 2 within a single retry loop. The 92 occurrences in 7d reflect retry-amplified observations.

**Fix:** Add `agent._cross_turn_overload_hits` — a cross-turn counter on the agent instance (like `_consecutive_stale_streams`). Increment on each 503/overloaded response, decay by 1 on non-overloaded responses. The `_should_fallback` gate now fires at `>= 3` cumulative hits, catching persistent overload that resets per turn.

### #1243 — Multi-shot refusal recovery with escalation
**Root cause:** The old code used `_refusal_nudged_this_turn` (boolean), firing one advisory nudge per turn. A repeated refusal after the nudge was silently accepted as the final answer — no escalation, no second chance.

**Fix:** Replace the boolean with `_refusal_nudge_count` (counter). 1st refusal: advisory nudge (unchanged). 2nd: escalated directive with stronger language. 3rd: accepted. `maybe_refusal_nudge` no longer returns `None` on `already_nudged=True` — the caller handles escalation language.

## Test plan
- [x] 262 existing tests pass (`test_tool_guardrails.py`, `test_loop_guard.py`, `test_tool_call_guardrail_runtime.py`, `test_read_patch_resilience.py`, `test_fuzzy_match.py`, `test_terminal_retry_spiral.py`)
- [x] 10 new tests in `test_evolution_2026_07_24.py` covering all 5 fixes
- [x] Updated `test_already_nudged_returns_none` → `test_already_nudged_still_detects` to match new multi-shot behavior
- [x] Updated `test_patch_replace_clean_message` to assert `success=True` instead of error message

Closes #1242, #974, #1205, #1243, #1258


Lexus2016:
⚠️ **Holding merge — independent review (external model) found a permanent-wedge regression in the #1242 portion.** The #1243 (refusal-nudge) and #1258 (patch spiral-cap + no-op-edit-not-failure) changes look sound; the #1242 change does not.

**Bug:** recording guardrail-**blocked** (never-executed) tool calls as failures via `after_call(failed=True)` (`agent/tool_executor.py` ~907-913 and ~1652-1658) permanently wedges a tool:
1. Once `_cross_turn_tool_failure_counts[tool]` hits the cap, `before_call()` blocks the call **pre-flight** (`agent/tool_guardrails.py:380-415`).
2. The only recovery is a **successful** `after_call(..., failed=False)` decaying the count (`tool_guardrails.py:644-659`) — but the call is blocked, so it never executes, so it never succeeds, so the count never decays.
3. The cap is keyed by `tool_name`, not signature (`tool_guardrails.py:380,532-533`), so a **corrected** call is blocked too. Concrete: `terminal` (or `patch`, now in `_SPIRAL_PRONE_TOOLS`) reaches cap → model corrects to a valid command → still blocked → count increments again → never recovers.

**Also:** the change's premise appears incorrect — the counter freezing at cap is *already* enough to re-fire across turns, because `reset_for_turn()` clears `_halt_decision` but NOT `_cross_turn_tool_failure_counts` (`tool_guardrails.py:346-350`) and `before_call()` re-blocks from the persisted count. Existing test `tests/agent/test_tool_guardrails.py:644-655` documents this. So the #1242 change is both harmful and unnecessary.

**Recommendation:** drop the `after_call(failed=True)`-on-blocked additions (keep #1243/#1258), or add a recovery path (decay the cross-turn count on a blocked-then-corrected call, or a bounded block-count that expires). CI didn't catch this because no test exercises "corrected call after cap must eventually recover."

Lexus2016:
Pushed a fix commit removing the #1242 blocked-call-as-failure change (the permanent-wedge regression above). The PR now lands only the sound parts: #1243 (refusal recovery), #1258 (patch spiral-cap + identical-edit success), and cross-turn overload counting. #1242 is left for a proper fix (recovery path). 86 tests green.

PR #1266, opened 2026-07-24
fix: provider recovery — billing fallback bypass, overload breaker logging, refusal session escalation

### LINKED ISSUE #1263: [FIX] Provider 503 overloaded errors persist at 92 occurrences in 7d — fix #1205 did not hold, no circuit breaker recovery
The agent repeatedly cannot recover from provider 503 overloaded errors. When the primary model backend returns HTTP 503, the agent retries or fails the task instead of cleanly falling back to an alternate provider or backing off and retrying with exponential delay.

- Frequency: 92 occurrences of 503 overloaded errors across the last 7 days (6938 sessions scanned)
- Tool/area involved: Provider layer / model backend HTTP transport
- Failure shape: HTTP 503 returned by provider on model inference call; agent does not invoke circuit breaker or fallback chain; errors counted but not acted upon

### Impact on real tasks
A 503 overload means the model backend is temporarily unavailable. Without a proper backoff+fallback path, every such error either fails the user task outright or forces the agent into unproductive retry spirals (up to 55-deep terminal retries observed alongside), wasting tokens and time while never recovering the actual task.

A prior fix (#1205) targeted this exact signal at 92 occurrences but the issue was closed and the signal persists at the same level — the fix did not hold.

### Proposed direction
Implement a hard circuit breaker at the provider layer that:
1. On ≥2 consecutive 503 responses, marks the provider as temporarily degraded
2. Backs off with exponential delay (e.g., 2s, 4s, 8s) before retrying the same provider
3. Falls back to the next provider in the configured chain on sustained 503
4. Logs the circuit breaker state so the agent can distinguish "backing off" from "permanently failed" and communicate appropriately to the user

The timeout circuit breaker (#1150) reduced timeouts from 637→116 using this pattern; the same mechanism should be applied to 503 overload responses.

### Value
- Impact: 0.9 (blocks tasks on every 503 occurrence; 92x in 7d is high frequency)
- Effort: 0.5 (pattern already proven by #1150 timeout circuit breaker; adapt for 503)
- Priority Score: 0.9 × 2 × (1 − 0.4 × 0.5) = 1.44

### LINKED ISSUE #1264: [FIX] Provider 403 billing errors persist at 63 occurrences in 7d — no provider fallback on quota exhaustion
The agent repeatedly cannot recover from provider 403 billing errors. When the model backend returns HTTP 403 (billing/quota exhausted), the agent retries or fails the task instead of cleanly falling back to an alternate provider.

- Frequency: 63 occurrences of 403 billing errors across the last 7 days (6938 sessions scanned)
- Tool/area involved: Provider layer / model backend HTTP transport (billing/quota path)
- Failure shape: HTTP 403 returned by provider on model inference call; agent does not invoke provider fallback chain; errors counted but not acted upon

### Impact on real tasks
A 403 billing error means the model backend rejects requests due to account quota or billing limits. Without a fallback to an alternate provider, every such error either fails the user task outright or forces the agent into unproductive retries, wasting tokens and time. 63 occurrences in 7 days indicates this is a recurring operational condition, not a one-off.

Prior fixes (#1043, #1231) targeted this signal but were closed; 403 billing errors continue at 63 occurrences — the fixes did not hold.

### Proposed direction
Implement a hard circuit breaker at the provider layer that:
1. On the first 403 billing response, marks the provider as quota-exhausted (not a transient error)
2. Immediately falls back to the next provider in the configured chain (no retry on same provider)
3. Emits a clear log/event so the agent can communicate to the user that the primary provider needs billing attention
4. Optionally retries the exhausted provider after a configurable cooldown (e.g., 1 hour) rather than hammering it

The timeout circuit breaker (#1150) and its adaptation for 503 (#1263) demonstrate this pattern; the same mechanism should be extended to 403 billing errors.

### Value
- Impact: 0.8 (blocks tasks on every 403 occurrence; 63x in 7d is significant)
- Effort: 0.4 (pattern already proven by #1150/#1263 circuit breakers; adapt for 403)
- Priority Score: 0.8 × 2 × (1 − 0.4 × 0.4) = 1.34

### LINKED ISSUE #1265: [FIX] Refusal/access-denied signal rises monotonically (865→973→1113→1635→1689) — taxonomy classifies but never prevents or recovers
The agent generates refusal or access-denied language at an extremely high and monotonically rising rate. The count has risen across every introspection cycle: 865 → 973 → 1113 → 1635 → 1689. The refusal taxonomy module classifies these messages but does not prevent or recover from them — the signal keeps growing.

- Frequency: 1689 refusal/access-denied signals across the last 7 days (6938 sessions scanned)
- Trend: monotonically rising across 5 consecutive introspection cycles (865 → 973 → 1113 → 1635 → 1689)
- Tool/area involved: refusal_taxonomy module (classifies but does not prevent)
- Failure shape: agent produces language indicating it cannot proceed (access denied, not supported, cannot do X) and the user's goal is left unmet; the taxonomy labels it but the agent does not change behavior

### Impact on real tasks
At 1689 occurrences in 7 days, this is the single highest-volume signal in the entire introspection digest — higher than all tool failures combined. Each occurrence represents a task where the agent told the user it could not proceed, and the taxonomy module merely categorizes the refusal rather than triggering a recovery path (retry with different approach, escalate, or suggest an alternative).

Prior fix #1243 (closed, needs-regression label) targeted this at 1635; the signal has risen to 1689 since then — the classification-only approach is not working.

### Proposed direction
Move from passive classification to active intervention:
1. When the refusal taxonomy detects a refusal pattern, inject a recovery directive into the agent context — e.g., "You refused this task. Consider: (a) using a different tool, (b) asking for clarification, (c) breaking the task into sub-tasks, (d) explaining what specifically is blocked."
2. Track refusal rate per session and trigger a strategy change (tool switch, context refresh, delegate to subagent) when it exceeds a threshold
3. Distinguish between genuine refusals (missing capability, safety constraint) and false refusals (agent unnecessarily cautious on a task it can do) — the latter should be auto-retried with a stronger "attempt the task" directive

### Value
- Impact: 1.0 (highest-volume signal; blocks tasks on every occurrence; monotonically worsening)
- Effort: 0.6 (requires wiring taxonomy output into agent recovery loop, not just logging)
- Priority Score: 1.0 × 2 × (1 − 0.4 × 0.6) = 1.52

Lexus2016:
## Summary

Three provider-recovery fixes from the 2026-07-24 evolution analysis:

### #1264 — Billing fallback bypass (403 billing, 63 occurrences)
Billing/quota exhaustion is account-level, not credential-level. When a multi-credential pool exists, the code was calling `_pool_may_recover_from_rate_limit()` which returns `True`, causing the fallback to be deferred in favor of credential rotation. But rotating credentials on the same exhausted account will also 403 — the quota is per-account. This bypasses the pool check for `FailoverReason.billing` so the fallback chain is attempted immediately.

### #1263 — Overload breaker status logging (503 overload, 92 occurrences)
When the overload circuit breaker trips (2 consecutive 503s or 3 cross-turn) but no fallback is activated (no chain configured, chain exhausted, or pool rotation preferred), the code fell through to backoff/retry silently. Now emits a clear status message so the agent/user can distinguish "backing off" from "permanently failed" and knows a fallback would help.

### #1265 — Per-session refusal escalation (refusal/access-denied, rising signal)
The per-turn refusal nudge (max 2 per turn) was insufficient when refusals kept recurring across turns. Added a session-level counter (`_session_refusal_count`) that tracks refusals across the entire session. When the threshold is met (per-turn nudges exhausted ≥ 2 AND session refusals ≥ 4), injects a strategy-change directive suggesting concrete alternatives: different tool, delegation, smaller steps, or explicit capability gap explanation.

Also adds `failure_reason` to the non-retryable handler return dict for consistency with other failure handlers (lines 4087, 4941 already had it).

## Test plan
- [x] `test_billing_error_no_fallback_fails` — 403 billing error fails fast (1 API call, no wasted retries)
- [x] `test_overload_breaker_emits_status_no_fallback` — 2× 503 triggers circuit breaker status message
- [x] `test_refusal_session_counter_pattern` — session counter uses getattr default 0, increments correctly
- [x] `test_refusal_escalation_threshold` — escalation fires only when per-turn ≥ 2 AND session ≥ 4
- [x] All 89 existing tests in `test_rate_limit_fail_fast.py` + `test_loop_guard.py` still pass

Closes #1263, Closes #1264, Closes #1265

### DIFF
--- a/agent/conversation_loop.py
+++ b/agent/conversation_loop.py
@@ -3848,9 +3848,19 @@ def _run_conversation_impl(
                     # pool can't help when the *upstream* model (DeepSeek,
                     # etc.) is throttling OpenRouter, so always fall back to a
                     # different model regardless of pool state.
+                    # #1264 — Billing/quota exhaustion is account-level, not
+                    # credential-level.  Rotating credentials on the same
+                    # exhausted account won't help — the quota is per-account,
+                    # not per-key.  Bypass the pool check for billing so the
+                    # fallback chain is always attempted immediately instead
+                    # of burning a rotation that will also 403.
                     _is_upstream = classified.reason == FailoverReason.upstream_rate_limit
+                    _skip_pool_check = (
+                        _is_upstream
+                        or classified.reason == FailoverReason.billing
+                    )
                     pool_may_recover = (
-                        False if _is_upstream
+                        False if _skip_pool_check
                         else _ra()._pool_may_recover_from_rate_limit(
                             agent._credential_pool,
                         )
@@ -3901,6 +3911,25 @@ def _run_conversation_impl(
                             # _consecutive_stale_streams reset on provider swap.
                             agent._cross_turn_overload_hits = 0
                             continue
+
+                # #1263 — When the overload breaker trips but no fallback is
+                # activated (no chain configured, chain exhausted, or pool
+                # rotation preferred), the code falls through to backoff/retry
+                # silently.  Emit a clear status so the agent/user can
+                # distinguish "backing off" from "permanently failed".
+                if (
+                    classified.reason == FailoverReason.overloaded
+                    and (
+                        _retry.consecutive_overload_hits >= 2
+                        or getattr(agent, "_cross_turn_overload_hits", 0) >= 3
+                    )
+                ):
+                    agent._buffer_status(
+                        "⚠️ Provider overloaded (503) — circuit breaker "
+                        "tripped, no fallback activated. Backing off with "
+                        "extended delay; the provider may recover. "
+                        "Configure a fallback provider to auto-recover."
+                    )
 
                 # ── Auth-failure provider failover ───────────────────────
                 # A 401/403 that survives the per-provider credential-refresh
@@ -4698,6 +4727,7 @@ def _run_conversation_impl(
                         "completed": False,
                         "failed": True,
                         "error": _nonretryable_summary,
+                        "failure_reason": classified.reason.value,
                     }
 
                 if retry_count >= max_retries:
@@ -6363,6 +6393,9 @@ def _run_conversation_impl(
                         _refusal_nudge = None
                     if _refusal_nudge:
                         agent._refusal_nudge_count = _refusal_count + 1
+                        agent._session_refusal_count = getattr(
+                            agent, "_session_refusal_count", 0
+                        ) + 1
                         # #1243 — Escalate the nudge language on the 2nd
                         # refusal to a directive rather than advisory.
                         if _refusal_count >= 1:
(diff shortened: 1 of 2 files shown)


Lexus2016:

Lexus2016:

Lexus2016:

PR #1284, opened 2026-07-25
feat: raise memory char limit from 2200 to 8000 (Closes #1283)

### LINKED ISSUE #1283: [UX] memory tool hard-rejects at 2,200-char capacity limit — forces manual garbage collection (3 failures, 5-deep spiral)
The built-in `memory` tool rejects entries when the total stored content exceeds a fixed character budget (currently 2,200 chars). When the agent has accumulated multiple notes and tries to add new information, the tool returns a capacity-limit error and instructs the agent to use `replace` to consolidate. This forces the agent to spend multiple extra turns reading existing memory, choosing what to evict, and rewriting — a manual garbage-collection burden.

- Frequency: 2 capacity-limit rejections in recent sessions (1 session had a 5-deep consecutive cluster of memory calls before recovery)
- Tool/area involved: `memory` (built-in)
- Failure shape: `"memory at N/2200 chars; adding this entry would exceed the limit; use 'replace' to consolidate"` — the agent must then manually merge/evict
- Recovery: agent recovered by performing a replace/consolidation on the next call, but this required reading all existing entries and rewriting

### Impact on real tasks
The fixed 2,200-char budget is reached quickly in any session that accumulates structured data. Each capacity hit costs 3-5 extra tool calls (read-all → decide-eviction → rewrite → retry). This makes the built-in memory impractical for anything beyond a few pointers, pushing users toward external stores unnecessarily.

### Proposed direction
- Increase the character budget (e.g. to 8,000-10,000), OR
- Implement automatic LRU/semantic eviction when capacity is reached (instead of hard-rejecting), OR
- Allow the agent to append with auto-truncation of oldest entries rather than forcing manual consolidation

### Value
- Impact: 0.5
- Effort: 0.2
- Priority Score: 0.92

Lexus2016:
Raises the default memory_char_limit from 2,200 to 8,000 chars across config.py, memory_tool.py, and memory_audit.py. Updates test assertion. Closes #1283. Automated evolution PR.

### DIFF
--- a/hermes_cli/config.py
+++ b/hermes_cli/config.py
@@ -2474,7 +2474,7 @@ DEFAULT_CONFIG = {
         #                     /memory reject <id>.
         # To disable memory entirely, use memory_enabled: false instead.
         "write_approval": False,
-        "memory_char_limit": 2200,  # ~800 tokens at 2.75 chars/token
+        "memory_char_limit": 8000,  # ~2,900 tokens at 2.75 chars/token
         "user_char_limit": 1375,  # ~500 tokens at 2.75 chars/token
         # External memory provider plugin (empty = built-in only).
         # Set to a provider name to activate: "openviking", "mem0",
--- a/skills/productivity/memory-audit/scripts/memory_audit.py
+++ b/skills/productivity/memory-audit/scripts/memory_audit.py
@@ -36,7 +36,7 @@ from typing import List, Optional, Tuple
 # is a standalone script (it runs via cron, not inside the agent), so it
 # reads no config. The caller can override via CLI flags.  If you changed
 # your memory limits in config.yaml, pass them explicitly.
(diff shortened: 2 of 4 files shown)


Lexus2016:
Closing as superseded: #1286 (evolution/issue-1283-memory-auto-evict) addresses #1283 with the same issue's preferred direction — automatic oldest-entry eviction instead of a hard reject — rather than just raising the budget cap. Both PRs touch overlapping files (config.py, memory_tool.py) and both carry `Closes #1283`, so only one can land. #1286 is the root-cause fix; this PR (raising 2200→8000) is a band-aid that merely delays the same capacity spiral. Leaving #1286 for human review (it trips the 200-line autonomous merge cap). Recycle the limit-raise idea only if #1286 is rejected.

PR #1291, opened 2026-07-26
feat: add temporal metadata (valid_from/valid_until/supersedes) to Mem0 entries (Closes #1289)

### LINKED ISSUE #1289: Add temporal metadata (valid_from/valid_until/supersedes) to memory entries to recover the ~32-point production accuracy gap — supersedes 2026-07-25 draft #2 (narrowed to minimal fix)
## Problem statement

The production memory fidelity gap is now **quantified**. New analysis (RankSquire, ranksquire.com/2026/05/06) concretizes what prior research only flagged qualitatively:

- **Mem0 v0.8.2 scores 91.6 on the LoCoMo benchmark but only ~49.0% effective accuracy in production after 30 days at a 38% staleness rate.**
- **Memory Fidelity Curve (RankSquire):** `Production Accuracy ≈ Benchmark − 0.22 × StalenessRate − 0.15 × log₁₀(Entities)`
- At 38% staleness, that's roughly **~32 accuracy points lost** to staleness alone.
- **Root cause:** Mem0 stores timestamps but has NO as-of queries and NO supersession semantics; stale facts and superseded facts are retrieved as if current.
- **Zep/Graphiti is the only system with native bi-temporal validity windows** (`valid_at`/`invalid_at`); Cognee is graph-native but NOT temporal-strategy-native.

**Key insight from today's research:** the staleness problem is recoverable largely from **temporal metadata + as-of query logic**, INDEPENDENT of backend choice. Adopting a full graph backend is a separate, larger decision; adding `valid_from`/`valid_until`/`supersedes` + as-of retrieval is the minimal, high-ROI fix.

## Proposed approach

1. **Minimal fix (this issue):** Add temporal metadata to every memory entry:
   - `valid_from` (when the fact became true)
   - `valid_until` (when it ceased to be true / was superseded; null = still current)
   - `supersedes` (id of the prior entry this one replaces)
2. **As-of retrieval:** Add an as-of query mode — retrieval filters to entries where `valid_from <= as_of < valid_until` (or `valid_until IS NULL`), excluding superseded entries.
3. **Supersession logic:** When a new fact contradicts an existing one, mark the old `valid_until = now` and link via `supersedes` rather than overwriting.
4. **Measure:** Re-run the RankSquire Memory Fidelity Curve against Hermes memory with and without the temporal fields; target recovering a meaningful share of the ~32-point gap.
5. **Defer (NOT this issue):** Full Graphiti-style entity-relationship graph backend. Track as optional follow-up pending the measurement above.

## Acceptance criteria

- [ ] Memory entry schema includes `valid_from`, `valid_until`, `supersedes` (nullable)
- [ ] As-of query mode implemented and tested (only non-stale, non-superseded entries returned)
- [ ] Supersession path implemented: contradicting facts mark the prior entry `valid_until`, link via `supersedes`
- [ ] Measurement run: Hermes-memory Memory Fidelity Curve before vs after, documented
- [ ] Measurable accuracy improvement on temporal/staleness queries (target: recover a meaningful share of the ~32-point gap)
- [ ] No regression on non-temporal retrieval

## References

- RankSquire Memory Fidelity analysis: https://ranksquire.com/2026/05/06/long-term-memory-for-ai-agents
- Mem0 / Zep / Graphiti / Cognee comparison: https://mnemoverse.com/docs/library/ai-memory-solutions-2026-q3
- Graphiti: https://github.com/getzep/graphiti (28,981 stars, Jul 20)
- getzep.com

## Source

Research brief 2026-07-26 — Vector #4 (Production memory fidelity gap quantified, YELLOW). Supersedes & narrows 2026-07-25 draft #2.

Lexus2016:
Automated evolution PR for issue #1289.

Layers bi-temporal validity on top of the existing Mem0 backend to recover the ~32-point production accuracy gap from staleness (RankSquire Memory Fidelity Curve). Backend-agnostic: operates purely on the metadata field each backend already forwards.

Validation: ruff check ✓, pytest tests/plugins/memory/ 500 passed (incl. 19 new temporal tests), diff 199 lines (≤ 200-line cap).

### DIFF
--- a/plugins/memory/mem0/__init__.py
+++ b/plugins/memory/mem0/__init__.py
@@ -38,7 +38,7 @@ import logging
 import os
 import threading
 import time
-from typing import Any, Dict, List
+from typing import Any, Dict, List, Optional
 
 from agent.memory_provider import MemoryProvider
 from tools.registry import tool_error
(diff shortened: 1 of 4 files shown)


Lexus2016:
Blocked by code review (self-audit + second-model review): TZ-correctness bug in `_parse_iso_epoch` defeats the feature on every non-UTC host.

**The bug:** `_parse_iso_epoch()` uses `time.mktime()` which interprets the parsed struct_time in LOCAL time, but `_now_iso()` writes UTC via `time.gmtime()`. Verified by execution under 3 timezones:
- UTC: 0s offset — correct (this is why CI passes).
- America/New_York (UTC-5): −14400s offset → `check >= ts_until` is False → **superseded/stale entries are NOT filtered out** (the feature silently no-ops).
- Asia/Tokyo (UTC+9): +32400s offset → ts_until shifts into the past → **legitimately-current entries wrongly excluded → recall shrinks**, violating the PR's own 'recall never shrinks' fail-open contract.

The bug does NOT trip the fail-open branch (timestamp is well-formed, returns a wrong-not-None epoch). Tests are structurally blind: they run TZ=UTC and feed `time.time()` on both write+compare sides, so the offset cancels.

**To fix (one line):** replace the body of `_parse_iso_epoch` with `calendar.timegm(time.strptime(value, "%Y-%m-%dT%H:%M:%SZ"))` (the format is always `Z`, so drop the `Z`→`+0000`/`%z` dance). Add a parametrized test under `TZ=America/New_York` and `TZ=Asia/Tokyo` asserting the round-trip is within ~2s of `time.time()` so this can't regress.

The idea is sound — rework this one narrow fix and re-open.

PR #1316, opened 2026-07-26
fix(hydra-gate): suppress false-positive wake-ups via dispatch ledger (Closes #1305)

Lexus2016:
## Summary

Automated evolution PR for issue #1305.

Adds a **dispatch-ledger gate predicate** to `scripts/evolution_hydra_gate.py`. A pipeline pair already adjudicated with a terminal verdict (`STEADY_STATE` / `CONSUMED`) today is no longer re-evaluated every cron tick — eliminating the 20+/day `integration→upstream-sync` false-positive wake-ups reported in the issue.

## Changes
- `_dispatch_ledger_path()` — canonical path to `hydra-dispatch-<date>.jsonl`.
- `_load_terminal_verdicts()` — reads today's terminal-verdict records (fails open: absent/malformed ledger → no suppression → wake).
- `_check_pool()` — a pair with a terminal verdict today is forced to `fresh=False`.
- 6 new tests covering: terminal-verdict suppression (STEADY_STATE + CONSUMED), non-terminal verdict passthrough, missing/malformed ledger, and end-to-end `main()` wake suppression.

## Design notes
- **Fail-open**: an unreadable or missing ledger never silently suppresses a real wake — the gate only *adds* suppression, never *removes* it.
- Accepts both `stage` and `edge` keys for the pair identifier (orchestrator record shape).
- Verdicts are matched against a frozen set `{STEADY_STATE, CONSUMED}` so only genuinely terminal adjudications suppress.

## Validation
- `ruff check` ✓
- `ruff format --check` ✓
- `pytest tests/scripts/test_evolution_hydra_gate.py` → 10 passed
- Diff: 175 insertions, 2 deletions (within the 200-line autonomous merge cap)

Closes #1305

PR #1318, opened 2026-07-26
fix(hydra-gate): self-recording dispatch ledger stops same-day false-positive wake-ups (#1305)

### LINKED ISSUE #1305: [BUG] Hydra-gate false-positive wake-ups: integration→upstream-sync edge fired 20+ times/day for no-ops
## Problem statement (introspection-derived)

The Hydra-gate fires false-positive wake-ups for the `integration → upstream-sync` edge. On 2026-07-26 alone it fired **20+ times**, each waking the orchestrator for a no-op. The gate currently uses raw file `mtime` instead of a dispatch ledger, so it re-evaluates and re-fires even after confirming `STEADY_STATE` / `CONSUMED`. Each skip still woke the orchestrator, dispatched the agent, and ran a full read cycle of 4 state surfaces — wasting agent API calls and generating large dispatch logs.

## Evidence

- `hydra-dispatch-2026-07-26.jsonl` contains 13 tick records (ticks 10–22); every one records `integration → upstream-sync` as a FALSE POSITIVE skip. Tick 22 reports *"20th FALSE POSITIVE"*.
- The Hydra's own `blockers` field says: *"META (CRITICAL): hydra-gate fired 18+ false-positive wake-ups today. RECOMMENDATION: add gate predicate."*

## Proposed fix

Add a gate predicate to the hydra-gate:

> `(stage_last_dispatched.date == today AND verdict in {STEADY_STATE, CONSUMED}) → suppress wake-up`

Replace the raw-mtime check with a dispatch ledger so a stage already adjudicated as `STEADY_STATE`/`CONSUMED` today is not re-evaluated. This is exactly the fix the Hydra's own tick entries recommend.

## Acceptance criteria

- [ ] A dispatch ledger (not raw mtime) gates wake-ups.
- [ ] A stage that resolved to `STEADY_STATE` or `CONSUMED` today does NOT wake the orchestrator again that day.
- [ ] On a following-day run, the 20+/day false-positive count drops to 0 (or only genuine state-change wake-ups fire).

## Source

- Introspection: `/config/.hermes/evolution/introspection/2026-07-26.json` (severity: **medium**)
- Example session: `cron_bc6edf39bede` (evolution-hydra) ticks 3–22 on 2026-07-26

Lexus2016:
Automated evolution PR for issue #1305.

## Problem
The Hydra gate gated wake-ups purely on raw upstream file mtime, with no record of whether it had already dispatched for an edge that day. A stage already adjudicated STEADY_STATE/CONSUMED was re-evaluated and re-fired on every tick — observed at 20+ wasted orchestrator wake-ups/day on the integration→upstream-sync edge alone, each spending agent API tokens and running a full read cycle for a no-op.

## Fix
Add a self-recording per-edge dispatch ledger (`~/.hermes/evolution/.hydra-dispatch-ledger.json`), keyed by edge + date, storing the upstream mtime at dispatch time. A subsequent same-day tick suppresses the edge unless the upstream produced STRICTLY newer output (genuine new material re-fires).

**Why self-recording, not orchestrator-fed:** a prior attempt (branch `evolution/issue-1305-hydra-dispatch-ledger`, abandoned) read an orchestrator-written `hydra-dispatch-<date>.jsonl`. But the gate runs BEFORE the orchestrator — that file is never written, so the suppression never fires. This version records its own dispatch decision at wake time, so it works without orchestrator cooperation.

Fail-open: a corrupt/missing ledger never blocks a real wake-up. Entries prune after 7 days.

## Tests
9 regression tests: ledger round-trip + fail-open, no-entry/stale-date/same-mtime/newer-mtime semantics, first-tick-records, second-tick-suppressed (the core #1305 regression), newer-upstream-re-fires, prune, corrupt-ledger.

## Size note
316 insertions / 6 deletions — exceeds the 200-line autonomous-merge cap because the gate-fix and its regression tests are one safety-critical unit (a buggy ledger could suppress a genuine wake-up, so the tests are load-bearing). Held for human merge.

Closes #1305

### DIFF
--- a/scripts/evolution_hydra_gate.py
+++ b/scripts/evolution_hydra_gate.py
@@ -12,6 +12,13 @@ The gate checks upstream→downstream staleness for the 7 evolution stages and
 returns false (sleep) when every consumer is ahead of or equal to its producer.
 It also sleeps when ``halt-state.txt`` is present, preventing expensive LLM work
 on a broken pipeline (#770).
+
+A per-edge dispatch ledger (#1305) suppresses false-positive wake-ups: once the
+gate has dispatched for an edge on a given day, subsequent ticks on the same day
+do NOT re-fire unless the upstream stage produced NEW output (a strictly newer
+mtime) since the last dispatch. Without this ledger the raw-mtime check re-fires
+on every tick — observed at 20+ wasted orchestrator wake-ups/day on the
+integration→upstream-sync edge alone (#1305).
 """
 
 from __future__ import annotations
@@ -22,7 +29,12 @@ import subprocess
 import sys
 from datetime import datetime, timedelta
 from pathlib import Path
-from typing import Dict, Tuple
+from typing import Dict, Optional, Tuple
+
+# Ledger entries older than this are pruned to keep the file bounded. Entries are
+# keyed by day, so a week of history is ample for debugging a "did the gate fire
+# today?" question without growing unbounded over months.
+_LEDGER_PRUNE_DAYS = 7
 
 
 def _hot_path() -> Path:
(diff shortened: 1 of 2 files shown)


Lexus2016:
Closing as duplicate of PR #1316, which fixes the same issue (#1305). #1316 is being merged because it is more faithful to the issue: it reads the existing `hydra-dispatch-<date>.jsonl` ledger that the orchestrator already writes (the evidence in #1305 explicitly cites this file) and suppresses edges with terminal verdicts (STEADY_STATE/CONSUMED), rather than introducing a parallel mtime-based ledger. This approach matches the proposed fix in the issue exactly.

PR #1319, opened 2026-07-26
fix(mcp): pre-flight guard stops turbo-memory-mcp retry storm (#1297)

### LINKED ISSUE #1297: [BUG] turbo-memory-mcp binary missing on PATH — root cause of persistent tqmemory retry loop (189 failures/scan)
## Source
Sourced from introspection manifest `introspection/2026-07-26.json` pattern `tqmemory-binary-missing-path` (severity HIGH, frequency 189). Not a duplicate — the prior `tqmemory-retry-loop` pattern tracked the symptom; this issue tracks the **root cause** now identified.

## Problem statement
The `tqmemory` MCP server fails to start because the executable `turbo-memory-mcp` is not found on `PATH`. The wrapper `/opt/hermes-evolution/tools/mcp_stdio_watchdog.py` raises `FileNotFoundError` at **line 121** when spawning the subprocess. This is NOT a transient network/timeout issue — the binary is **absent from the environment**. Every session and cron run that initializes MCP triggers 3 retry attempts × multiple revival cycles, producing the 189-frequency retry-loop pattern observed in the logs.

## Impact
- **Frequency:** 189 occurrences in a single introspection scan (14 sessions)
- **Effect:** drowns real errors in errors.log, wastes MCP initialization budget on every run, masks other MCP failures
- **User-visible:** tqmemory memory backend effectively unusable — silently degrades to no-op

## Proposed fix (introspection-recommended)
1. **Install** the `turbo-memory-mcp` binary into the runtime PATH (or document it as a required dependency in setup-hermes.sh).
2. **Pre-flight check:** before spawning, verify the binary exists on PATH and emit a single clean error (not 3× retries × N cycles) when it is missing — e.g. add an `if not shutil.which("turbo-memory-mcp"):` guard at the watchdog spawn site that raises a single actionable error instead of letting FileNotFoundError bubble through the retry loop.

## Effort
Trivial (effort 0.1, < 1 hour): one dependency install + a small pre-flight guard. Zero behavioral risk — the server is already failing; this only changes the failure mode from noisy-retry to clean-error, and (when the binary is installed) makes it actually work.

## Acceptance criteria
- [ ] `turbo-memory-mcp` resolves on PATH in the runtime environment, OR
- [ ] watchdog emits a single actionable error when the binary is absent (no retry storm)
- [ ] introspection scan frequency for this pattern drops from 189 → 0 on the next cycle

Lexus2016:
Automated evolution PR for issue #1297.

## Problem
A stdio MCP server whose command binary is absent (e.g. `turbo-memory-mcp` missing from PATH) failed at `execvp` time inside the MCP SDK's `stdio_client` with `FileNotFoundError`. The reconnect loop treats that as a retryable connection error and re-attempts it `_MAX_INITIAL_CONNECT_RETRIES` (3) times, then parks and revives — producing **189 identical failure/scan** in errors.log (the #1297 retry storm) and drowning real errors.

## Fix
Add `_ensure_stdio_command_resolvable()`: a pre-flight in `_run_stdio`, run right after `_resolve_stdio_command`, that raises a **non-retryable** `MissingMcpCommandError` when the command is a bare name not on the subprocess PATH or an absolute path that does not exist.

`MissingMcpCommandError` subclasses `NonMcpEndpointError`, which the reconnect loop already catches-and-exits on cleanly (`except NonMcpEndpointError` → `self._ready.set(); return`). So the server fails **ONCE** with an actionable install hint instead of storming.

`_resolve_stdio_command` only does PATH/dir-candidate fallback for `npx/npm/node`; every other missing binary now hits this guard before any spawn is attempted.

## Design notes
- Resolves against the **subprocess env's PATH** (what the child actually sees), not the Hermes process PATH — a binary on the process PATH but absent from the subprocess PATH still fails correctly.
- Fails fast with a message naming the command + an install hint (e.g. `uv tool install git+https://github.com/Lexus2016/turbo_quant_memory` for turbo-memory-mcp).
- Zero behavioral risk: the server is already failing; this only changes the failure mode from a noisy 3×N retry storm to a single clean error.

## Tests
8 tests in `test_mcp_tool_issue_948.py`: missing bare command, bad absolute path, non-executable, empty command, healthy pass-through (bare-on-path + absolute-executable), the non-retryable subclass invariant, and subprocess-vs-process PATH scoping.

Closes #1297

### DIFF
--- a/tests/tools/test_mcp_tool_issue_948.py
+++ b/tests/tools/test_mcp_tool_issue_948.py
@@ -219,3 +219,125 @@ def test_run_stdio_malware_check_times_out_fail_open():
         assert elapsed < 1.0, f"startup did not fail-open promptly ({elapsed:.1f}s)"
 
     asyncio.run(_test())
+
+
+# ---------------------------------------------------------------------------
+# #1297: pre-flight guard — a missing stdio command must fail ONCE with a
+# non-retryable error, not bubble FileNotFoundError through the reconnect
+# retry loop (3 retries × N revival cycles = the observed 189-failure/scan
+# storm). The guard is exercised directly here; the integration (it stops the
+# retry loop because MissingMcpCommandError subclasses NonMcpEndpointError,
+# which the reconnect loop catches-and-exits on) is covered by inspection of
+# the loop's `except NonMcpEndpointError` branch.
+# ---------------------------------------------------------------------------
+
+import pytest  # noqa: E402
+
+from tools.mcp_tool import (  # noqa: E402
+    MissingMcpCommandError,
+    _ensure_stdio_command_resolvable,
+)
+
+
+def test_preflight_raises_for_missing_bare_command():
+    """A bare command name not on PATH → MissingMcpCommandError (non-retryable),
+    naming the command and giving an install hint. This is the turbo-memory-mcp
+    case from #1297."""
+    with patch("tools.mcp_tool.shutil.which", return_value=None):
+        with pytest.raises(MissingMcpCommandError) as exc_info:
+            _ensure_stdio_command_resolvable(
+                "tqmemory", "turbo-memory-mcp", {"PATH": "/usr/bin"}
+            )
+    msg = str(exc_info.value)
+    assert "turbo-memory-mcp" in msg
+    assert "not found on PATH" in msg
+    # Actionable install hint must be present.
+    assert "install" in msg.lower()
+
+
+def test_preflight_raises_for_nonexistent_absolute_path():
+    """An absolute path that does not exist → MissingMcpCommandError with a
+    'path does not exist' message (distinct from the PATH case)."""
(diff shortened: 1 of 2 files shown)


Lexus2016:
Closing: pre-flight guard breaks an existing test (test_hanging_initialize_is_bounded_not_leaked in test_mcp_stdio_init_timeout.py). The new `_ensure_stdio_command_resolvable` raises `MissingMcpCommandError` BEFORE the test's timeout logic can engage — the test uses a fake 'fake-mcp' binary to exercise the timeout path, but the pre-flight rejects it as non-existent. The fix needs to either: (1) mock `_ensure_stdio_command_resolvable` in that test, or (2) restructure the pre-flight so it can be bypassed in test fixtures that intentionally use non-resolvable commands to test other behavior. Sent back for rework.

PR #1329, opened 2026-07-26
fix(tools): patch pre-flight distinguishes missing-file from generic read failure (Closes #1326)

### LINKED ISSUE #1326: [FIX] patch tool 24% failure rate — dominant class is 'target file not found' (68%), not fuzzy-match
The `patch` tool fails on ~24% of calls, but the **dominant error class is "target file not found / unreadable"** (68% of patch failures), NOT fuzzy-match / whitespace / indentation issues that prior fixes targeted.

- Frequency: 25 genuine errors out of 104 patch calls across recent sessions (7-day window, 221 sessions scanned)
- Tool/area involved: `patch` (replace mode)
- Failure shape breakdown:
  - "target file not found": 17/25 (68%) — patch aimed at a path that does not exist or cannot be read
  - "failed to read target file": 2/25
  - "empty old_string argument": 2/25
  - "old_string fuzzy match failed": 2/25
  - "old_string not unique": 2/25
- Prior fixes (#889, #976, #1037, #1258) targeted fuzzy-match / whitespace / indentation patterns — those are now near-zero. The signal **regressed to a new dominant class** that was never addressed: the agent attempts to edit files before confirming they exist.

### Impact on real tasks
Every patch-against-missing-file wastes 2-3 tool calls (failed patch → read_file to check → retry or give up). When the file genuinely doesn't exist, the agent often loops trying alternate paths instead of surfacing "this file does not exist; should I create it?" The 68% file-not-found rate means the agent is editing blind — attempting edits without first verifying the target.

### Proposed direction
1. **Pre-flight existence check**: before applying a patch in `replace` mode, the tool should verify the file exists and is readable; if not, return a distinct error ("file does not exist — use write_file to create, or confirm the path") instead of a generic match failure.
2. **Create-on-miss hint**: when the target file is missing, suggest `write_file` as the correct tool, since `patch` is for editing existing files.
3. **Agent-side guard**: the system prompt should instruct the agent to confirm a file exists (via read_file or search_files) before issuing a patch, especially for paths it hasn't read in the current session.

### Value
- Impact: 0.75 (24% failure rate on a core editing tool, recurring across many sessions, blocks code changes)
- Effort: 0.3 (pre-flight stat() check is a small, localized change)
- Priority Score: 0.75 × 2 × (1 − 0.4 × 0.3) = 1.17

Lexus2016:
## Summary

Addresses **#1326** — the `patch` tool's 24% failure rate was dominated (68%) by "target file not found", but the agent couldn't tell, so it kept retrying `patch` against non-existent paths instead of switching to `write_file`.

## Root cause

`patch_replace` read the target via `cat <path> 2>/dev/null`. The `2>/dev/null` **discarded the reason** the read failed, so a missing file, a directory, and a permission error all collapsed to the identical generic `"Failed to read file: {path}"`. That string maps to `error_class="read_error"` via `classify_file_error` — **never** `not_found` — so the recovery hint never steered the agent toward `write_file` (the correct tool to create a missing file).

## Fix

On a failed read (happy path unchanged — the probes run **only** on failure), `patch_replace` now calls `_diagnose_read_failure()`, which distinguishes:

| Read-failure reason | Error produced | `classify_file_error` class | Recovery hint |
|---|---|---|---|
| File missing | reuses `_suggest_similar_files` (similar files + nearest ancestor) | `not_found` | "Create it with `write_file`…" |
| Path is a directory | `Cannot patch a directory: <path> is a directory` | (not `read_error`) | — |
| File unreadable (perms) | `Permission denied reading file: <path>…` | — | check permissions |

This makes the dominant failure class self-correcting: the agent now receives a `not_found` classification with a `write_file` recovery hint, exactly as the issue requested. The v4a patch path was already correct (it routes through `read_file_raw` → `_suggest_similar_files`), so only `patch_replace` needed the fix.

## Verification

- **3 new tests** (`TestPatchReplaceReadFailureClassification`) pin the classification: missing-file → `not_found` + `write_file` hint; directory → clear error; permission → distinct message.
- **187 regression tests** green (`test_file_operations.py`, `test_patch_self_correction.py`, `test_read_patch_resilience.py`, `test_file_write_safety.py`).
- **Pre-PR test shard**: PASSED.
- **Lint** (`ruff check`): clean.
- **Diff**: 148 lines (under the 200-line autonomous merge cap).

Closes #1326

🤖 Automated evolution PR.

### DIFF
--- a/tests/tools/test_file_operations.py
+++ b/tests/tools/test_file_operations.py
@@ -903,3 +903,101 @@ class _DeletedTestGitBaselineCheck:
     helper is restored or replaced.
     """
     pass
+
+
+# =========================================================================
+# #1326 — patch_replace file-not-found preflight
+# =========================================================================
+# Before #1326, ``cat <path> 2>/dev/null`` swallowed WHY the read failed, so
+# every miss (missing file / directory / permission error) collapsed to the
+# identical generic "Failed to read file: {path}". That string classifies as
+# ``read_error`` (never ``not_found``), so the agent never learned that
+# ``write_file`` — not ``patch`` — is the right tool to create the missing
+# file. File-not-found was the dominant patch-failure class (68% of errors).
+# These tests pin the new distinct-classification behavior.
+
+
+class TestPatchReplaceReadFailureClassification:
+    """patch_replace must distinguish missing-file / directory / permission
+    read failures and surface a create-on-miss hint, not a generic error."""
+
+    def test_missing_file_classifies_as_not_found(self, mock_env):
+        """A patch against a non-existent path must produce an error that
+        ``classify_file_error`` maps to ``not_found`` (with a write_file
+        recovery hint), NOT the old generic ``read_error``."""
+        target = "/tmp/does_not_exist_xyz_1326.py"
+
+        def side_effect(command, **kwargs):
+            # The initial ``cat <path> 2>/dev/null`` fails (file absent).
+            if command.startswith("cat ") and target in command:
+                return {"output": "", "returncode": 1}
+            # _diagnose_read_failure: ``test -e <path>`` → no (file absent).
+            if command.startswith("test -e ") and target in command:
+                return {"output": "no\n", "returncode": 0}
+            # _suggest_similar_files lists the parent dir — empty here.
+            if command.startswith("ls -1 "):
+                return {"output": "", "returncode": 1}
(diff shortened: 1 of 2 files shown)


Lexus2016:

Lexus2016:

Lexus2016:

Issue #1325, opened 2026-07-26
[FIX] Introspection digest attributes failures by tool but not by reason — fix→regress→refile treadmill on 4 tools


Lexus2016:
## Problem (observed in real usage)
The introspection digest attributes failures by tool but does not sub-classify failure *reason*, so recurring-symptom issues get filed and "fixed" repeatedly while the distinct root causes (file-not-found vs read-above-limit vs non-JSON envelope, for the same tool) are never separated. Ten+ circuit-breaker fixes have landed and the headline counts keep climbing.

### Evidence (aggregated, anonymized)
- In the current 7-day digest: `terminal` 1961 failures / 660 sessions, `read_file` 773 / 220 sessions, `patch` 287 / 24 sessions, `tool_call` 298 / 42 sessions, `search_files` 246 / 249 sessions. These are aggregates of multiple distinct failure modes lumped into one counter.
- The realized-impact ledger shows the same tools being "fixed" and re-regressing in a loop: terminal (#602→#694→#737→#942→#1022→#974→#1167), read_file (#782→#886→#970→#1044→#1092→#1120), patch (#697→#889→#976→#1037→#1258). Each fix closes the aggregate issue; each subsequent digest re-triggers because a *different* sub-cause is still firing.
- The extractor (`introspection_extract.py`) classifies a result as failed via `_tool_result_failed` reading the structured status — correct — but then attributes it only to the tool name, discarding the `error`/`exit_code`/message that would distinguish the sub-cause. There is no `failure_reason` breakdown.

### Impact on real tasks
Fixes that target one failure mode are recorded as fixing the whole tool, so the issue closes and the other modes continue unchecked until the next digest re-opens a near-identical issue. This produces a steady stream of duplicate work (the closed-issues list is dominated by re-fixes of the same 4 tools) and means the user's real blockers underneath those tools are never fully cleared — each "fix" buys days, not resolution.

### Proposed direction
Add a lightweight failure-reason sub-classifier to the extractor (no LLM — keyword/field rules on the structured `error`/`exit_code`/status):
1. Define a small taxonomy per tool family: `file-not-found`, `permission-denied`, `timeout`, `non-zero-exit`, `parse-error`, `no-match` (patch), `quota/billing` (provider), `other`.
2. Emit `tool_failures_by_reason: {tool: {reason: count}}` in the digest alongside the existing aggregate.
[...]

PR #1199, opened 2026-07-21
fix(#1188): browser_navigate spiral cap regression — 15-deep spirals

Lexus2016:
## Summary

Fixes #1188 — `browser_navigate` failures regress with 15-deep consecutive spirals across 30 sessions, despite the existing cap=3.

## Root Cause

Two independent root causes prevented the cap from enforcing:

### 1. `bot_detection_warning` misclassified as success (`agent/display.py`)

`browser_navigate` returns `{"success": true, "bot_detection_warning": "..."}` when it successfully navigates to a bot-detection/Cloudflare/captcha page (the browser-level navigation succeeded, but the page is a blocked page). The failure classifier `_detect_tool_failure` saw `success: true` and returned `failed=False` — so the guardrail counter never incremented. The model retried 15× against a blocked page that will never load properly.

**Fix:** Added a `bot_detection_warning` check to both `_detect_tool_failure` (production path) and `classify_tool_failure` (fallback classifier) — when `success=true` AND `bot_detection_warning` is present, classify as `[bot detection]` failure.

### 2. Intermittent success resets cross-turn streak (`agent/tool_guardrails.py`)

A single successful navigation (e.g., to a different, fast-loading page between failing attempts) fully cleared `_cross_turn_tool_failure_counts` via `pop()`, resetting the streak to 0. The cap never accumulated because the pattern `fail, succeed, fail, succeed, fail...` kept the counter below the threshold.

**Fix:** Decay the cross-turn streak by 1 on success instead of clearing it. Multiple consecutive successes still drain to 0 (a genuinely recovered backend is not penalized), but a fail/succeed/fail pattern now accumulates toward the cap and eventually halts.

## Verification

- E2E reproduction confirmed both root causes (standalone scripts, not committed)
- `scripts/run_tests.sh tests/agent/test_tool_guardrails.py` — 65/65 tests pass
- `scripts/run_tests.sh tests/agent/test_display.py tests/tools/test_browser_navigate_fallback.py tests/agent/test_tool_result_classification.py` — 75/75 pass
- 4 new tests + 2 updated tests covering both fixes

## Files Changed

- `agent/display.py` — `_detect_tool_failure`: classify `bot_detection_warning` as failure
- `agent/tool_guardrails.py` — `classify_tool_failure`: mirror the bot_detection check; `after_call`: decay cross-turn streak by 1 on success instead of clearing
- `tests/agent/test_tool_guardrails.py` — 4 new + 2 updated tests

PR #1189, opened 2026-07-20
fix(guardrails): cap deferred-tool + memory spirals (#1185, #1186, #1187)

Lexus2016:
## Summary

Extends the existing always-on spiral-prone-tool failure cap to cover the whole deferred-tool loading chain (`tool_search` → `tool_describe` → `tool_call`) plus `memory` operations. These three tools had **no circuit breaker** — the agent blind-retried the same failing call up to 13 consecutive times with no fallback directive, while the core tools (`terminal`, `execute_code`, `read_file`, `process`, `search_files`) were already capped.

This is the established, tested mechanism (`_SPIRAL_PRONE_TOOLS` + `_TOOL_FALLBACK_DIRECTIVE` in `agent/tool_guardrails.py`): consecutive same-tool failures hit the `spiral_failure_cap` (default 5) and halt the turn with a concrete fallback directive, regardless of `hard_stop_enabled`. Adding these three tool names is a 2-line-per-tool change that takes effect through the existing `before_call`/`after_call` guards — no new dispatch-site code, no new surface.

## Evidence (trace-miner, aggregated/anonymized, 7-day window)

- **#1185 — `tool_call`**: 168 failures / 21 sessions, 13-deep consecutive spirals. The agent loads a deferred tool via search/describe/call, the call fails (wrong args schema, tool unavailable, provider rejection), and it retries the identical call repeatedly instead of switching strategy.
- **#1186 — `memory`**: 94 failures / 21 sessions, 11-deep consecutive retries. Regressed ~9x from the 10x that triggered #1135/#1136 (those fixes closed completed but the signal climbed back). A single corrupted/locked store causes a long spiral that never recovers.
- **#1187 — `tool_describe`**: 59 failures, the middle step of the search→describe→call chain. When it fails (tool not in catalog, schema fetch error, stale name), the agent has no schema to invoke the deferred tool and either re-describes the same name or falls back to re-running `tool_search` (which itself spirals).

## Changes

**`agent/tool_guardrails.py`** (2 edits):
1. `_SPIRAL_PRONE_TOOLS` — added `tool_call`, `tool_describe`, `memory` alongside the existing five, with a comment citing #1185/#1186/#1187.
2. `_TOOL_FALLBACK_DIRECTIVE` — added actionable directives for each:
   - `tool_call` → run `tool_search` for an alternative, use `tool_describe` to validate schema, or invoke a core tool instead of blind-retrying.
[...]

Issue #1167, opened 2026-07-18
[FIX] Terminal retry spiral regressed to 1600 failures with 55-deep spirals across 557 sessions


Lexus2016:
### Problem Statement
The terminal tool's retry spiral has **regressed again** despite 9+ prior closed fixes (#365, #231, #737, #942, #974, #1022, #1091, #1109, #1122, #1124). The deterministic trace miner reports over the last 7 days:

- **`terminal` tool failures: 1600 occurrences in 7d** (up from #1022's 1346 and #974's 1237).
- **`terminal` retry spiral: up to 55 consecutive retries across 557 sessions** (up from #1022's 55-deep across 474 sessions and #942's 404 sessions).

This is a demonstrated, recurring, and worsening problem. Each prior fix landed and the signal either stayed flat or got worse within 2-7 days of the merge (see closed #1140 — "Merged spiral/failure fixes do not reduce the live signal"). The tool wrapper or its preconditions are not being hardened at the failure-class level; the fixes appear to address individual symptoms while the underlying non-retryable-failure classification + fallback path is still not catching the bulk of cases.

### Proposed Solution
Treat this as a **bug-class fix, not a single-site patch**. The recurring pattern across 9 prior fixes indicates the current approach (per-fix hardening) is insufficient. The fix should:

1. **Classify terminal failures into retryable vs non-retryable at the wrapper level**, so that non-retryable failures (e.g., a command that exits non-zero with a deterministic error — permission denied, command not found, syntax error) do NOT trigger the retry loop at all. Only genuinely transient failures (timeout under load, brief I/O hiccup) should retry.
2. **Add a hard execution-level circuit breaker** for the retry spiral: a per-session consecutive-retry cap that, when hit, breaks the spiral by returning a structured non-retryable diagnostic to the model instead of allowing another retry. (Prior issue #1034 proposed a soft guardrail that was systematically ignored.)
3. **Surface a fallback path**: when the terminal tool fails non-retryably, suggest an alternative tool (e.g., `read_file` for file reads, `search_files` for greps, `execute_code` for multi-step composition) in the diagnostic, so the model can switch tools instead of retrying the same failing call.

The goal is not to make the terminal tool never fail — it is to make failures **non-retryable by default** and to **break the spiral hard** when it starts, with a deterministic diagnostic that points the model at the right next action.

### Value Proposition
[...]

PR #1332, opened 2026-07-27
fix(#1324,#1325): normalize failure rate by session volume + classify failure reason

### LINKED ISSUE #1324: [FIX] Realized-impact verdicts produce false regressions — raw counts not normalized by growing session volume
Post-merge realized-impact verdicts systematically produce false "regressed" results because they compare raw failure counts across digest windows without normalizing for session volume — which is growing rapidly on this install.

- The introspection extract reported 7212 sessions scanned in this cycle, up from ~4860 a few cycles prior (~48% growth) and ~2602 earlier. Raw failure counts grow roughly proportionally, but the per-session failure rate (the actual signal) held steady or improved.
- The realized-impact ledger currently holds 42 verdicts: 13 `regressed`, 20 `no-signal`, 9 `confirmed`. At least 4 recent `regressed` verdicts (#1189, #1199, #970, #974) were recorded against tools whose per-session failure rate was flat or down — the raw count rose only because the session denominator rose.
- Concrete example: tool_call went 168 failures / 21 sessions (8.0/session) → 298 / 42 sessions (**7.1/session — improved**), max-consecutive unchanged at 13. Recorded as `regressed` because 298 > 168.
- The `record-verdict` helper accepts a free-text note but the signal the operator reads is the headline count comparison. There is no session-normalized rate in the digest (`tool_failures` is an absolute count, `sessions_scanned` is available but never divided).

### Impact on real tasks
False `regressed` verdicts trigger the close-loop gate to re-open issues that are actually holding, piling `needs-regression` work onto the implementation pipeline and burning cycles re-investigating fixes that are working. It also erodes trust in the verification loop: if half the verdicts are false alarms, maintainers stop acting on the real regressions. This directly degrades goal 3 (verify merges actually helped) — the measurement itself is miscalibrated.

### Proposed direction
Normalize the failure signal by session volume before any comparison:
1. In `introspection_extract.py`, emit `failure_rate` = `tool_failures[tool] / sessions_with_tool_failures` (or per total sessions) alongside the raw count, and `spiral_depth_per_session` = max_consecutive normalized by sessions.
2. In `evolution_realized_impact.py` `record-verdict` / `check-close`, compare the per-session rate (delta in rate), not the raw count. A verdict should flip to `regressed` only if the rate increased beyond a threshold (e.g. >20%), not merely because the denominator grew.
3. Store the baseline rate at merge time in the ledger so the comparison is apples-to-apples.

### Value
- Impact: 0.9 (the entire realized-impact loop's correctness depends on this; currently producing noise that wastes implementation cycles and re-opens working fixes)
- Effort: 0.3 (arithmetic in two scripts + a stored baseline field)
- Priority Score: 0.9 * 2 * (1 - 0.4 * 0.3) = **1.0**

Lexus2016:
## Summary

Fixes #1324 and #1325 — two introspection-signal quality bugs that caused false regression verdicts and a fix→regress→refile treadmill.

### #1324 — False regression verdicts from unnormalized failure counts

**Problem:** Realized-impact verdicts produced false regressions because raw `tool_failures` counts weren't normalized by session volume. As session volume grows cycle-over-cycle (e.g. 21→42 sessions), the raw count rises proportionally (168→298) even when per-session rate holds or improves (8.0→7.1/session) — so a raw-count comparison flips a stable/improving tool to "regressed". This produced false-regression verdicts on at least 4 issues (#1189, #1199, #970, #974).

**Fix:**
- `introspection_extract.py`: emit `tool_failure_rate` (failures / sessions_scanned) alongside raw `tool_failures` (raw counts retained for back-compat)
- `evolution_realized_impact.py`: `record_merge()` now accepts optional `baseline_tool_failure_rate` to snapshot the rate at merge time
- New `compare_failure_rate()` pure function: regressed = rate increased >20%, improved = rate dropped >5%, stable = within ±threshold
- New `compare-rate` CLI subcommand for deterministic rate-based verdicts
- `evolution-integration` SKILL.md: record baseline rate at merge time
- `evolution-introspection` SKILL.md: compare rates not raw counts, use `compare-rate` helper

### #1325 — Fix→regress→refile treadmill from missing failure-reason attribution

**Problem:** The digest attributed failures by tool but not by reason, so a fix that addressed one failure mode (file-not-found on `read_file`) could not be distinguished from a regression in another mode (timeout) on the same tool. This caused fix chains: terminal (#602→#694→#737→#942→#1022→#974→#1167), read_file (#782→#886→#970→#1044→#1092→#1120), patch (#697→#889→#976→#1037→#1258).

**Fix:**
- `introspection_extract.py`: reuse `ErrorClassifier` from `evolution/lib/root_cause_diagnosis.py` (keyword-based, no LLM) to sub-classify each failed result
- Emit `tool_failures_by_reason: {tool: {reason: count}}` in the digest (reasons: not_found, permission, timeout, resource_limit, validation, syntax_error, network, other)
- `evolution-introspection` SKILL.md: file and verify at the reason level, surface top reason in issue titles, don't re-file original when a different reason rose

## Test plan

- [x] 58 tests pass (22 new for `evolution_realized_impact`, 7 new for `tool_failures_by_reason`, 29 existing introspection_extract still green)
- [x] `compare_failure_rate` pure function: improved/regressed/stable/new/gone verdicts, zero-baseline edge case, custom thresholds, None inputs
- [x] `record_merge` with/without baseline, baseline survives verdict folding
- [x] CLI `compare-rate` and `record-merge` subcommands with baseline arg
- [x] `tool_failures_by_reason` classification: not_found, permission, timeout, multiple reasons same tool, no failures, other bucket, cross-session aggregation

## Size note

725 insertions / 293 deletions across 6 files. Exceeds 200-line self-merge cap — **held for human merge**. Production code is ~273 lines; the bulk is tests (680 lines).

### DIFF
--- a/scripts/evolution_realized_impact.py
+++ b/scripts/evolution_realized_impact.py
@@ -90,17 +90,104 @@ def record_merge(
     merged_at: str,
     predicted_impact: float,
     target: str,
+    baseline_tool_failure_rate: Optional[Dict[str, float]] = None,
 ) -> None:
-    """Record a merge event in the realized-impact ledger."""
-    append_ledger_record(
-        ledger_file,
+    """Record a merge event in the realized-impact ledger.
+
+    ``baseline_tool_failure_rate`` (optional, #1324) snapshots the per-tool
+    failure rate (= failures / sessions_scanned) at merge time so the
+    post-merge verification can compare rates instead of raw counts. Raw counts
+    grow with session volume, so a raw-count comparison flips a stable/improving
+    tool to "regressed" purely because more sessions were scanned. The rate
+    normalizes that out.
+    """
+    rec: Dict[str, Any] = {
+        "issue": issue,
+        "merged_at": merged_at,
+        "predicted_impact": predicted_impact,
+        "target": target,
+    }
+    if baseline_tool_failure_rate is not None:
+        rec["baseline_tool_failure_rate"] = baseline_tool_failure_rate
+    append_ledger_record(ledger_file, rec)
+
+
+# ── Rate-based regression comparison (#1324) ──────────────────────────────
+# The raw ``tool_failures`` count grows with session volume: 168 failures over
+# 21 sessions (8.0/session) becomes 298 over 42 sessions (7.1/session = BETTER)
+# but a raw-count comparison calls it "regressed" because 298 > 168. Comparing
+# per-session rates instead eliminates this false regression. A tool is
+# "regressed" only if its rate increased by more than ``regression_threshold``
+# (default 20% — absorbs natural variance without masking real regressions).
+
+REGRESSION_THRESHOLD = 0.20  # rate must increase >20% to count as regressed
+IMPROVEMENT_THRESHOLD = 0.05  # rate must drop >5% to count as improved (lower bar — any meaningful drop is real progress)
+
+
+def compare_failure_rate(
+    baseline: Optional[Dict[str, float]],
+    current: Optional[Dict[str, float]],
+    regression_threshold: float = REGRESSION_THRESHOLD,
+    improvement_threshold: float = IMPROVEMENT_THRESHOLD,
+) -> Dict[str, Any]:
+    """Compare per-tool failure rates between baseline and current digest.
+
+    Returns a dict with per-tool verdicts and an overall signal::
+
         {
-            "issue": issue,
-            "merged_at": merged_at,
-            "predicted_impact": predicted_impact,
-            "target": target,
-        },
-    )
+          "tools": {"terminal": {"baseline": 8.0, "current": 7.1, "verdict": "improved"}, ...},
+          "any_regressed": False,
+          "regressed_tools": [],
+          "improved_tools": ["terminal"],
+        }
+
+    A tool is ``"regressed"`` only if its current rate exceeds baseline by more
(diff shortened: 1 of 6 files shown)


Lexus2016:
Closing: half of this PR is already in `main`, and the other half shipped in #1336.

- **The #1324 half** (failure-rate normalization) was merged on 2026-07-27 as ddd0d28c6 via #1334, which added `failure_rate`, `baseline_failure_rate` and `classify_verdict_by_rate`. Issue #1324 is closed as completed. This overlap is the direct cause of the merge conflict here — two implementations of the same normalization competing in `introspection_extract.py` and `evolution_realized_impact.py`.
- **The #1325 half** (failure-reason sub-classification) shipped in #1336, merged as 032e47058, with an equivalent `tool_failures_by_reason` emission and a wider reason taxonomy.

Rebasing this branch would mean hand-merging a redundant normalization against the merged one, for no behaviour that is not already on `main`.

**One piece here is not yet on `main`** and is worth refiling on its own: the SKILL.md updates to `evolution-integration` (record the baseline rate at merge time) and `evolution-introspection` (compare rates, not raw counts). The merged #1334 changed the scripts but the skill instructions were not updated to match, so the stages may still be documented against raw counts. Filed as a follow-up rather than carried in this conflicted branch.

Issue #1366, opened 2026-07-27
[INCREMENT] Refusal-recovery rate tracking (Child B of #1327)


Lexus2016:
## Parent: #1327 (soft-capability refusals plateau at 51/week)

**Depends on:** #1356 (Child A — recovery-before-refusal prompt nudge), merged as `bb0a325cc`.

### Problem

Child A added `RECOVERY_BEFORE_REFUSAL_GUIDANCE` to the stable system-prompt tier: before refusing for a missing capability, the agent should check whether an existing tool reaches the same goal another way, and propose that instead.

**Nothing measures whether it worked.** The extractor counts `refusals` with a single regex over assistant text (`_REFUSAL_RE` in `scripts/introspection_extract.py`) and cannot distinguish:

- a **bare refusal** — "I can't do that, there's no tool for it" (the failure #1327 is about), from
- a **recovered refusal** — "I can't do that directly, but I can achieve the same result via `terminal`" (exactly what Child A asks for)

Both increment the same counter. So the 51/week signal cannot move in the digest even if Child A works perfectly, and a future realized-impact verdict on #1356 has no rate to compare against — it would be forced to `no-signal`.

### Scope (~100 lines)

1. Add a conservative recovery-marker detector to `introspection_extract.py` — an assistant turn that both refuses AND proposes an alternative path counts as recovered, not bare.
2. Emit alongside the existing aggregate (additive, back-compat preserved):
   - `refusals_with_recovery` — count
   - `refusal_recovery_rate` — recovered ÷ total refusals
3. Keep `refusals` / `refusals_or_access_denied` meaning exactly what they mean today.
4. No LLM — keyword rules over assistant text, same discipline as the #1325 reason taxonomy.

### Success criteria

- [ ] A refusal that proposes an alternative is counted as recovered
- [ ] A bare refusal is not
- [ ] `refusal_recovery_rate` appears in the digest
- [ ] Existing `refusals` counts are unchanged

### Why it matters

This is the measurement half of #1327. Without it Child A is unfalsifiable, and Child C (recovery dispatcher, ~150 lines) would be built with no way to tell whether it improved anything.

---
*Child B of #1327. Child C (recovery dispatcher logic) still unfiled.*

Issue #1327, opened 2026-07-26
[UX] Soft-capability refusals plateau at 51/week — agent declines instead of proposing alternatives


Lexus2016:
## Problem (observed in real usage)
The agent frequently **declines tasks by citing a missing capability** ("I can't do X — no tool for Y", "model doesn't support Z") instead of attempting an alternative path. This is a soft-refusal pattern distinct from hard permission denials, and it leaves the user's goal unmet.

### Evidence (aggregated, anonymized)
- Frequency: 51 soft-capability declinations + 18 hard permission/access denials = 69 total refusal-shaped signals in recent sessions (7-day window, 221 sessions scanned; deterministic extractor reports 61 refusals_or_access_denied)
- Tool/area involved: no single tool — these are agent utterance patterns ("I can't", "not supported", "no access", "cannot proceed")
- Failure shape: 74% are soft-capability declinations (agent claims it lacks a tool/model/feature), 26% are hard permission denials
- Prior fixes (#975, #1035, #1243, #1265) built a refusal **taxonomy** that classifies these signals — but classification has never reduced the count. The signal dropped from its peak (1689) but plateaued at ~60/session-week, suggesting the taxonomy measures but does not change agent behavior.

### Impact on real tasks
When the agent says "I can't do X because I don't have tool Y," it stops. The user must then manually suggest the alternative the agent should have proposed itself (e.g., "use tool Z instead", "try this web search", "write a script"). This converts a self-service agent into a high-friction one, especially for tasks at the edge of the agent's stated capabilities where a creative alternative exists.

### Proposed direction
1. **Recovery-before-refusal protocol**: before emitting a refusal, the agent must propose at least one alternative approach (different tool, decomposition, web lookup, script-based workaround). The refusal taxonomy should feed a **recovery dispatcher** (issue #1027 proposed this but was closed without the dispatch behavior landing).
2. **Capability-edge nudge in system prompt**: add explicit guidance that "I don't have tool X" is not a terminal state — the agent should attempt the closest available substitute and report what it did.
3. **Track refusal-recovery rate**: extend the refusal taxonomy to measure how often a refusal was followed by a successful alternative, not just count refusals.

### Value
- Impact: 0.7 (51 soft-refusals/week leave user goals unmet; high friction)
- Effort: 0.5 (requires prompt engineering + recovery-logic + taxonomy extension)
- Priority Score: 0.7 × 2 × (1 − 0.4 × 0.5) = 1.12

[...]

Issue #1312, opened 2026-07-26
[IMPROVEMENT] Add a near-miss-vs-dead-end discriminator (metacognition) before re-issuing the same fix


Lexus2016:
### Problem Statement
A documented frontier agent failure mode (LHTB, Long-Horizon Terminal-Bench) is the **infinite debugging loop**: the agent cannot tell a near-miss (one step from success, keep pushing) from a dead end (re-applying the same fix = circling), so it loops until the time budget dies. Hermes has an `attention-reset` trigger on "stuck 2+ times", but that does not distinguish *a genuinely new attempt* from *the same argument re-issued* — so the same dead-end fix can cycle through the reset without ever being interrupted as a repeat.

### Proposed Solution
A small metacognition hook: before re-issuing a previously-tried fix (identical tool + identical/overlapping arguments) within N turns, the agent must produce a structured `{what_changed_since_last_attempt, why_expect_different_outcome}` check. If the diff is empty (same argument, same context), **force a strategy change** rather than a retry.

This is the same trigger-detection discipline as `attention-reset`, extended from "stuck 2+ times" to "re-issued identical argument within window" — a specific, narrow policy addition, not a generic "be smarter" ask.

### Value Proposition
- **Impact**: Medium (0.6)
- **Effort**: Medium (0.5)
- **Priority Score**: 0.96

### Research Evidence
- [Long-Horizon Terminal-Bench (LHTB)](https://zli12321.github.io/LHTB) — arXiv:2607.08964. The named second failure mode (after state-loss) is **infinite debugging loops**: agent cannot tell near-miss from dead-end, loops until the 90-min clock dies.

### Implementation Plan
1. Track a short rolling window (last N turns) of `{tool_name, normalized_args}` tuples per session
2. On a tool call whose `(tool, normalized_args)` matches a prior entry in the window, intercept and require a `{what_changed, why_expect_different}` structured justification
3. If justification is empty/trivial → block the retry and force a strategy-change prompt (mirrors `attention-reset` prompt structure)
4. Add a metric: `repeated_fix_attempts_per_session` to the evolution funnel — surfaces the dead-end class when it occurs
5. Tune N and the normalization (whitespace/quote/flag-order insensitive) on real traces

### Success Criteria
- [ ] Repeated identical-argument tool calls within the window trigger a justification prompt
- [ ] `repeated_fix_attempts_per_session` metric lands in the evolution funnel
[...]

Issue #1362, opened 2026-07-27
[FIX] Evolution SKILL.md files still document raw failure counts after #1324 normalization landed


Lexus2016:
## Problem

#1334 (merged as `ddd0d28c6`) fixed the false-regression bug by normalizing realized-impact verdicts by session volume. It added to the **scripts**:

- `scripts/introspection_extract.py` — emits `failure_rate` and `spiral_depth_per_session`
- `scripts/evolution_realized_impact.py` — `baseline_failure_rate` on `record_merge()`, plus `classify_verdict_by_rate()`

The **stage instructions were not updated to match**. Verified against current `main`:

- `skills/evolution/evolution-integration/SKILL.md` — no mention of `failure_rate`, `baseline_failure_rate`, or recording a baseline rate at merge time
- `skills/evolution/evolution-introspection/SKILL.md` — line 54 still instructs working from `tool_failures` (the raw per-tool count)

## Why it matters

The verdict logic can now compare rates, but the agent running these stages is still told to read raw counts. As session volume grows the raw count rises even when the per-session rate holds or improves — which is exactly the false-regression treadmill #1324 set out to kill. The fix is only half-deployed: correct primitives, stale instructions.

## Proposed fix

1. `evolution-integration/SKILL.md`: instruct recording `baseline_failure_rate` at merge time via `record_merge()`.
2. `evolution-introspection/SKILL.md`: instruct comparing `failure_rate` (not `tool_failures`) across digest windows, using `classify_verdict_by_rate()`.
3. Also worth surfacing `tool_failures_by_reason` (added by #1336, `032e47058`) so the digest attributes failures by reason, not just by tool.

## Provenance

Salvaged from PR #1332, which bundled this SKILL.md work with a competing implementation of the #1324 normalization. That PR was closed as redundant with #1334 + #1336; this documentation half is the one piece of it not otherwise on `main`.

PR #1374, opened 2026-07-27
fix: two review defects — dead-end detector silent on alternating calls, recovery metric counted deflection

Lexus2016:
Two defects found in post-merge review of work merged earlier today. Both are in code I wrote or repaired, both are confirmed by reproduction, and both would have degraded the signal the evolution loop runs on.

Independently flagged by a cross-provider review (Gemini via consilium) and then reproduced locally before fixing.

---

## 1. Dead-end detector went silent on alternating calls — `agent/loop_guard.py`

Repairing #1349 I added a liveness requirement: the repeated signature must be re-issued by the latest tool turn. That part is right — `5x terminal` then a `read_file`, or `8x terminal` then a text reply, count as progress everywhere else in that module (`TestRunBoundaries`).

**What was wrong is the order of operations.** The code ranked the window with `most_common(1)` and *then* required that winner to be live. When the agent alternates between two stuck calls, those two conditions select different signatures:

```
pytest, ruff, pytest, ruff, pytest, ruff, pytest, ruff   (each x4)
  most_common(1) -> "pytest"   (insertion order breaks the tie)
  live signature -> "ruff check"
  -> no match -> SILENT
```

Both calls are four-deep in identical re-issues. That is exactly the dead end #1312 exists to catch, and it went unreported.

**Fix:** look up the live signature first, gate on *its own* count. Liveness requirement unchanged.

New `TestAlternatingCallsAreStillDeadEnds` covers: fires at threshold, quiet below it, and names the live call rather than the most-common one.

---

## 2. Refusal-recovery counted deflection and explanation — `scripts/introspection_extract.py`

The pivot clause in `_RECOVERY_RE` allowed second person. So handing the work back to the human counted as a recovery:

| Text | Was counted | Actually |
|---|---|---|
| "I can't access that URL, however **you can** try visiting it yourself." | recovered | deflection |
| "I cannot do that. **Instead, please contact support**." | recovered | deflection |
| "I don't have permission. But **you will** need to run it **manually**." | recovered | deflection |
| "I can't do that, but **I can tell you why** it failed." | recovered | explanation, not an alternative path |

This inverts the metric: it improves whenever the agent gets *more polite about giving up* — the precise behaviour #1327 exists to eliminate.

**Fix:** pivot clause is first-person only, plus a `_NON_RECOVERY_RE` veto for deflection (`contact support`, `you can`, `yourself`, `manually`) and explanation-instead-of-action (`I can tell you why`). Vetoing beats narrowing the recovery pattern — these shapes are open-ended, and under-counting a real recovery is far cheaper than a rate that rises when the agent gives up nicely. Also adds first-person phrasings the original missed (`let me try`, `I'll write`).

### Measured impact on the real corpus

Same 4629 sessions, same 143 refusals:

```
before:  51 recovered   rate 0.3566
after:   31 recovered   rate 0.2168
```

**20 of the 51 were false positives — the headline number was inflated by ~64%.** The 21.7% figure is the one Child C (#1327 recovery dispatcher) should be measured against.

---

107 passing across the four loop-guard suites; 50 in `test_introspection_extract.py` (7 new). Ruff clean on all four files.

Refs #1312, #1366.

### DIFF
--- a/agent/loop_guard.py
+++ b/agent/loop_guard.py
@@ -512,28 +512,33 @@ def detect_dead_end_loop(
             signatures[key] += 1
             sig_to_tool[key] = name
 
-    # Find the most-repeated identical call.
     if not signatures:
         return None
-    top_key, top_count = signatures.most_common(1)[0]
+
+    # Gate on the signature the agent is issuing RIGHT NOW, and on ITS OWN count.
+    #
+    # Two separate requirements, and both matter:
+    #
+    # * Liveness — counting across the whole window would flag a run the agent
+    #   has already moved on from: 5x terminal followed by a read_file, or 8x
+    #   terminal followed by a text reply, are treated as progress everywhere
+    #   else in this module (see TestRunBoundaries). A turn calling several
+    #   DIFFERENT tools at once is varied work, so it does not count either.
+    #
+    # * Its own count, NOT the window's most-repeated one. Ranking by
+    #   ``most_common`` and then requiring that winner to be live silently
+    #   disabled the detector whenever the agent alternated between two stuck
+    #   calls: with `pytest` and `ruff` each re-issued four times, the counter
+    #   returns `pytest` (insertion order breaks the tie), the latest turn is
+    #   `ruff`, the two do not match, and a textbook dead end went unreported.
+    key = _last_call_signature(scan)
+    if key is None:
+        return None
+    top_count = signatures.get(key, 0)
     if top_count < _DEAD_END_REPEAT_THRESHOLD:
         return None
 
-    # The repeated signature must be what the agent is doing RIGHT NOW.
-    #
-    # Counting across the whole window would flag a run the agent has already
-    # moved on from: 5x terminal followed by a read_file, or 8x terminal
-    # followed by a text reply, are treated as progress everywhere else in this
-    # module (see TestRunBoundaries) — the agent broke the pattern by itself and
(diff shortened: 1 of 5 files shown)


Lexus2016:
Third and fourth defects found while fact-checking my own docs from #1362 against the code they describe. Both now in this PR (commit `21b98af99`).

**3. A threshold that does not exist.** The introspection SKILL.md claimed `classify_verdict_by_rate()` implements *"a rate rise >20% is regressed, a drop >5% is confirmed, anything between is stable"*. Reading the function:

```python
rel_delta = (current_rate - baseline_rate) / baseline_rate
return "regressed" if rel_delta > regression_threshold else "confirmed"
```

There is no 5% branch and no `stable` verdict. Three ledger values exist, not four. A skill file is executable guidance — an agent looking for a stable band that isn't there records the wrong verdict, and a wrong `regressed` re-opens issues that were actually fixed.

**4. Two smaller inaccuracies in the same block:**
- `failure_rate` was described as "failures ÷ sessions_scanned", implying a scalar. It is a dict keyed by tool, like `tool_failures`.
- The reason bucket is emitted as `quota/billing`; the doc listed `quota`, so an exact-match lookup finds nothing.

**Verified good:** the `record-merge` example. Rather than counting positionals by eye I ran the documented command and read the ledger line back — `0.42` lands in `baseline_failure_rate`. Unchanged.

Running total for this verification round: **4 defects, all in code or docs I merged today**, all reproduced before fixing.

PR #1380, opened 2026-07-27
feat: ensure leaf subagents retain terminal tool (Closes #1369)

### LINKED ISSUE #1369: [UX][REGRESSION] Delegated leaf subagents still arrive without terminal tool — 85 "I have no shell" spirals in 7d despite #648 fix

Delegated leaf subagents assigned shell/git/gh-dependent tasks (e.g. evolution pipeline steps, repo operations, build verification) repeatedly discover mid-task that they have no `terminal`/shell tool, then spiral through 5–21 "I have no shell/bash tool, I cannot proceed" turns before returning a failed result. The parent agent wastes a full delegation cycle (spin-up + context transfer + return) before learning the subagent was never equipped to do the work.

- Frequency: 85 refusal utterances of the "no shell/terminal/bash tool" shape across 20+ sessions in the last 7 days (441 sessions scanned)
- Tool/area involved: `delegate_task` / leaf-subagent toolset routing
- Failure shape: parent delegates a task whose completion requires shell access → subagent receives a toolset that omits `terminal` → subagent emits "I have no shell tool" and either loops, partially completes, or returns a blocked result → parent must re-dispatch or abandon
- Source: deterministic introspection extractor + targeted DB sampling (anonymized reason buckets only — no raw transcript text)

### Why this is a regression of #648 (not a duplicate)
#648 ("Delegated subagents lack required toolsets") was closed on 2026-07-01 with a toolset-inheritance fix. The current signal shows the fix did not hold for the `leaf` role specifically:
- The delegation tool defines `DEFAULT_TOOLSETS = ["terminal", "file", "web"]` and a leaf-role blocklist that removes `execute_code` (correct — children should reason, not script).
- However, in practice leaf subagents are still arriving without `terminal`, indicating the default-toolset path is not always applied (likely when the caller passes an explicit partial toolset, or when a profile/config narrows the inheritance), and there is no pre-dispatch validation that the requested task is compatible with the resolved toolset.

### Impact on real tasks
This blocks any autonomous workflow that delegates shell-dependent work (evolution pipeline stages, repo/git operations, build & test verification, MCP status checks). For cron-driven cycles the failure is silent and expensive — a full subagent spin-up is paid before the gap is discovered, and the parent often lacks the signal to retry with the correct toolset.

### Proposed direction
1. **Pre-dispatch toolset/task compatibility check**: when `delegate_task` is called, statically inspect the task goal/context for shell-dependent verbs (git, gh, build, test, run, shell, bash, install) and assert that `terminal` is in the resolved toolset; warn or auto-add it if missing.
2. **Make `DEFAULT_TOOLSETS` apply even when a partial explicit toolset is passed** (merge rather than replace), or document the override semantics clearly.
3. **Surface a structured "tool-missing" signal** (not a free-text refusal) so the parent agent can react with a corrected re-dispatch instead of reading a paragraph of subagent apology.

### Value
- Impact: 0.85 (silent blocker for autonomous cron + recurring across 20+ sessions/week)
- Effort: 0.35 (static verb-match on dispatch + toolset-merge fix + structured signal)
- Priority Score: 0.85 × 2 × (1 − 0.4 × 0.35) = 1.224

Lexus2016:
Merge terminal back into leaf children when explicit toolsets drop it and parent had it. Fixes 85 'I have no shell' spirals in 7d. Closes #1369

### DIFF
--- a/tests/tools/test_delegate_leaf_terminal.py
+++ b/tests/tools/test_delegate_leaf_terminal.py
@@ -1,0 +1,114 @@
+"""Tests for leaf subagent terminal tool retention (#1369).
+
+When explicit toolsets are given to delegate_task, the intersection with
+the parent's toolsets can drop 'terminal' even though the parent has it.
+Leaf subagents then spiral with "I have no shell" refusals. The fix
+merges 'terminal' back into the child's toolsets when the parent had it
+and the child role is 'leaf'.
+"""
+
+from types import SimpleNamespace
+
+import pytest
+
+from tools.delegate_tool import _strip_blocked_tools, _expand_parent_toolsets
+
+
+class TestLeafTerminalRetention:
+    """Verify terminal is merged back for leaf children when parent has it."""
+
+    def _resolve_child_toolsets(
+        self,
+        parent_enabled,
+        requested_toolsets,
+        effective_role="leaf",
+        parent_disabled=None,
+    ):
+        """Simulate the toolset resolution from _build_child_agent."""
+        parent = SimpleNamespace(
+            enabled_toolsets=parent_enabled,
+            disabled_toolsets=parent_disabled,
+            valid_tool_names=None,
+        )
+
+        # Derive parent toolsets
+        parent_enabled_val = getattr(parent, "enabled_toolsets", None)
+        if parent_enabled_val is not None:
+            parent_toolsets = set(parent_enabled_val)
+        else:
+            parent_toolsets = {"terminal", "file", "web"}
+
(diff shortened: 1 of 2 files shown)


Lexus2016:
Closed as duplicate: PR #1378 addresses the same issue (#1369) with a fuller implementation — a static goal/context heuristic (`_goal_needs_terminal`) with word-boundary regex detection plus a structured `toolset_adjusted` signal, vs. this PR's unconditional merge-back. Keeping the stronger one.

PR #1419, opened 2026-07-28
fix: validate delegated sub-agent resolves to ≥1 tool before launch (#1387)

### LINKED ISSUE #1387: [BUG] Validate delegated sub-agent tool lists resolve to ≥1 tool before launch
## Problem

When a delegated sub-agent's `tools` list resolves to an empty set (e.g., specified toolsets don't exist on the platform, or tool names are mistyped), the sub-agent launches with **no tools** and immediately spirals — repeating "I have no shell" or similar refusals until the turn limit is hit.

Claude Code fixed this exact bug in their July 2026 release: the Agent tool now returns a clear error naming the unrecognized entries instead of silently launching a toolless agent.

## Existing partial fix

PR #1378 / #1380 (Closes #1369) addressed the **terminal tool** specifically — auto-adding `terminal` when the goal needs shell access. But this is a targeted heuristic, not a general validation: a sub-agent could still be dispatched with an empty toolset if the specified tools are all unrecognized for reasons other than terminal-tool narrowing.

## Proposed fix

Add a **pre-dispatch validation** in `delegate_tool.py` (in `_build_child_agent` or equivalent) that:

1. After toolset resolution, checks whether the resolved toolset is empty or below a minimum threshold.
2. If empty: **fail fast** with a structured error naming the unrecognized tool/toolset entries, instead of launching a toolless sub-agent.
3. Include the unrecognized names in the error message so the parent agent can correct the delegation request.

This is the general case of #1369 — #1369 handles terminal-tool specifically; this handles the general "no tools resolved" case.

## Research source

Research brief 2026-07-28, item #12 — Claude Code July 2026 bug fixes (subagent tool resolution).
Source: https://releasebot.io/updates/anthropic/claude-code

## Labels

bug, delegation

Lexus2016:
Closes #1387

## Problem
When a delegated sub-agent's toolsets all resolved to nothing (unrecognized names, stripped by parent intersection, blocked tools), the child launched with ZERO tools and spiraled repeating 'I have no shell' until the turn limit.

## Fix
Added pre-dispatch validation in `delegate_task` after child construction but before execution: checks `child.valid_tool_names` and fails fast with a structured error naming the unresolved toolsets if the set is empty.

## Changes
- `tools/delegate_tool.py`: +30 lines (validation gate + denied-toolset stash)
- `tests/tools/test_delegate_empty_toolset_validation.py`: +102 lines (2 tests)
- Total: 132 insertions — within 200-line merge cap

## Validation
- ruff check ✓
- pytest: 2/2 passed ✓

### DIFF
--- a/tests/tools/test_delegate_empty_toolset_validation.py
+++ b/tests/tools/test_delegate_empty_toolset_validation.py
@@ -1,0 +1,102 @@
+"""Tests for delegate_tool pre-dispatch tool validation (#1387).
+
+Verifies that when a sub-agent's resolved toolset is empty (all requested
+toolsets were unrecognized or stripped), the delegation fails fast with a
+clear error instead of launching a toolless agent that spirals.
+"""
+
+import json
+from types import SimpleNamespace
+from unittest.mock import MagicMock, patch
+
+from tools.delegate_tool import delegate_task
+
+
+def _make_parent():
+    """Create a minimal mock parent agent."""
+    return SimpleNamespace(
+        enabled_toolsets=["terminal", "file"],
+        valid_tool_names=["terminal", "read_file", "write_file", "patch"],
+        model="test-model",
+        provider=None,
+        base_url=None,
+        api_key=None,
+        api_mode=None,
+        acp_command=None,
+        acp_args=[],
+        reasoning_config=None,
+        _fallback_chain=None,
+        providers_allowed=None,
+        providers_ignored=None,
+        providers_order=None,
+        provider_sort=None,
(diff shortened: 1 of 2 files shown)


Lexus2016:
Blocked by code-review: CI failing — the validation gate aborts the entire delegation with {'error': ...} instead of returning {'results': [...]} with a per-task error entry. This breaks the delegate_task result contract: 9+ tests in test_delegate.py expect 'results' key and fail with KeyError. The validation must return a per-task error WITHIN the results array, not abort the whole call.

PR #1421, opened 2026-07-28
fix: validate delegated sub-agent tool lists resolve to >=1 tool (Closes #1387)

### LINKED ISSUE #1387: [BUG] Validate delegated sub-agent tool lists resolve to ≥1 tool before launch

Lexus2016:
Automated evolution PR for issue #1387.

## Problem

When a delegated sub-agent's toolset resolution produced an empty list (requested toolsets don't intersect with parent's, or inherited toolsets are all stripped by _strip_blocked_tools), the sub-agent was launched with zero tools. This caused the sub-agent to spiral on every tool call — the root cause of the 'is not a deferrable tool' pattern observed in introspection (64 failed sessions, 4.14% rate).

## Fix

Added a validation guard in `_build_child_agent` after toolset resolution:
- If `child_toolsets` is empty AND toolsets were explicitly requested → `ValueError` naming the requested toolsets and parent's available toolsets
- If `child_toolsets` is empty from inherited toolsets → `ValueError` explaining the parent has no surviving enabled toolsets
- The caller in `delegate_task` catches the `ValueError` and returns a per-task error result with `status: "error"` instead of launching the toolless sub-agent

## Changes

- `tools/delegate_tool.py`: Empty-toolset guard in `_build_child_agent` (raises `ValueError`); try/except in `delegate_task` batch loop catches it and returns structured error result
- `tests/tools/test_delegate_toolset_scope.py`: Added `TestEmptyToolsetValidation` with 3 tests covering explicit-request-empty, inherited-empty, and valid-toolsets-no-raise

## Validation

- Lint: ✓ (ruff check)
- Tests: 14/14 passed (`tests/tools/test_delegate_toolset_scope.py`)

### DIFF
--- a/tests/tools/test_delegate_toolset_scope.py
+++ b/tests/tools/test_delegate_toolset_scope.py
@@ -6,6 +6,7 @@ parent's enabled_toolsets, it can escalate privileges by requesting
 arbitrary toolsets.
 """
 
+import pytest
 from types import SimpleNamespace
 
 from tools.delegate_tool import _strip_blocked_tools, _emit_parent_console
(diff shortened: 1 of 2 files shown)


Lexus2016:
Blocked by code-review: CI failing — the validation guard in _build_child_agent causes IndexError at delegate_tool.py:3240, breaking 8 tests in test_delegate_shallow_retry.py. The guard raises ValueError which is caught per-task, but the return-path change at line 3240 accesses an index that no longer exists in the modified control flow. Needs: fix the IndexError so existing shallow-retry tests pass, or restructure the guard to not alter the function's return-path shape.

Issue #102, opened 2026-06-10
[FIX] Delegate subagents frequently return empty results after ~3s, forcing parent redo


Lexus2016:
## Problem (observed in real usage)
`delegate_task` subagents often complete in ~3 seconds with a superficial summary and no extracted data, even though the parent requested a specific read/filter/compute operation. The parent agent must then re-run the same work inline, negating the purpose of delegation.

### Evidence (aggregated, anonymized)
- Frequency: seen in 5 of 25 recent sessions
- Tool/area involved: `delegate_task`, subagent terminal / file reads
- Failure shape: subagent reports "completed" but returns only a generic narrative; requested file contents, search results, or computed values are missing

### Impact on real tasks
Bulky read/filter tasks that should be offloaded to subagents instead consume extra parent turns, increasing token cost and slowing task completion.

### Proposed direction
Require subagents to echo key requested artifacts (file paths, counts, extracted snippets) explicitly in their summary. Add a validation step where the parent checks for emptiness and can auto-retry with stricter instructions.

### Value
- Impact: 0.7
- Effort: 0.5
- Priority Score: 2.8

Issue #1436, opened 2026-07-28
[FEATURE] Experience Memory Graph (EMG): graph-matched one-shot error correction from paired success/failure trajectories


Lexus2016:
## Summary

EMG (Experience Memory Graph) reformulates agent failure recovery as a **graph matching problem**. It converts completed `{failed_trajectory, successful_trajectory}` pairs into directed action-decision graphs, extracts common subgraphs (proven successful workflows) and graph edit paths (which actions to add/delete/relabel to go from failure to success), and stores them as a persistent memory graph. At inference time, the agent retrieves the best-matching graph edit path for the current task and executes it **in a single, loop-free pass** — no trial-and-error reflection cycle.

This is qualitatively different from the ERL proposal (#1303 / children #1359-#1361): ERL extracts *single-paragraph heuristics* from one trajectory; EMG extracts *structural diff operations* from paired trajectories and applies them as executable patches to the action plan. The two compose — ERL's heuristics inform *which* strategy to try; EMG's edit paths tell the agent *exactly which actions to change*. For autonomous real work where retries are costly or irreversible (cron jobs, production deployments), one-shot correction from accumulated experience is a step-change improvement over reflection-loop recovery.

## Source

- **Paper**: arXiv:2607.13884 — "Experience Memory Graph: One-Shot Error Correction for Agents" (Wang et al., Jul 15 2026)
- **Link**: https://arxiv.org/abs/2607.13884
- **Frontier standing**: behind
- **Related in-repo**: ERL pipeline (#1303, #1359, #1360, #1361) — EMG composes with, not replaces, ERL.

## Implementation Path for Hermes

The evolution pipeline already records `{task, trajectory, outcome}` tuples per cycle — a second extraction pass that converts paired success/failure trajectories into graph-edit-path entries, plus a retrieval hook at task start, turns existing telemetry into runtime error-prevention.

1. **Trajectory-pairing pass** (`scripts/evolution_emg_extract.py`): for tasks that have both a failed and a successful trajectory (same task descriptor, different outcomes), convert each trajectory into a directed action-decision graph (nodes = tool/action calls + decision points; edges = control flow). Compute the graph edit path between the failure and success graphs (node add/delete/relabel operations).
2. **Memory graph store**: persist graph-edit-path entries to `~/.hermes/evolution/emg/{date}.jsonl`, keyed by task descriptor signature, with intra-task nodes and cross-task edges.
[...]

Issue #1359, opened 2026-07-27
[FEATURE] ERL heuristic extraction from cross-task traces (Child A of #1303)


Lexus2016:
## Parent: #1303

Slice A of the ERL-style cross-task heuristic extraction pipeline.

## Scope (~150 lines)

Extract reusable heuristics from recorded agent task trajectories:
- Parse trajectory JSON files from `~/.hermes/evolution/trajectories/`
- Identify recurring (success_pattern, failure_pattern) pairs across tasks
- Score each heuristic by cross-task frequency + outcome correlation
- Write extracted heuristics to `~/.hermes/evolution/heuristics/{date}.json`

## Implementation
- New file: `scripts/evolution_heuristic_extract.py`
- Tests: `tests/scripts/test_evolution_heuristic_extract.py`
- Foundation for Slice B (judge-reranked retrieval) and Slice C (system-prompt injection)

## Acceptance Criteria
- Extraction pass runs over recorded trajectories and produces a ranked heuristic list
- Each heuristic has: pattern text, source trajectory IDs, frequency, outcome score
- Tests cover: empty trajectory set, single trajectory, multi-trajectory with shared patterns

## Merge cap
≤ 200 lines total (scripts + tests).

Lexus2016:
Blocked: dead code — `evolution_heuristic_extract.py` has zero call sites outside its own module and tests. PR #1397 was closed.

**What was wrong:** The extractor is a standalone module with no integration point — no cron job, no script import, no pipeline stage invokes it. It produces heuristics that nothing reads. This is the exact dead-code pattern PR #49 had.

**Definition of done:** Wire the extractor into the evolution pipeline so it actually runs:
1. Register it as a scheduled cron job (via `register_evolution_cron.py`) OR call it from an existing evolution stage so `run()` is invoked on real trajectory data.
2. Ensure the output heuristics (`~/.hermes/evolution/heuristics/{date}.json`) are consumed by Slice B/C or at minimum logged and retrievable.
3. Keep total diff ≤ 200 lines (the module is currently 235 lines alone — consider trimming `_RECOMMENDATIONS` or the CLI/argparse block, or splitting across multiple PRs if needed).

Re-open a PR once the extractor is invoked from the pipeline.

Lexus2016:
Blocked by #1363 — the input directory this slice parses is empty, and structurally so.

[...]

Issue #1270, opened 2026-07-25
[IMPROVEMENT] Selective memory addition gate with history-based utility deletion (Experience-Following Property)


Lexus2016:
### Problem Statement
The evolution pipeline appends every research report, every issues file, every implementation log indiscriminately — an **add-all** memory policy. The ACL 2026 memory-management paper (aclanthology.org/2026.acl-long.27) proves this self-degrades via three mechanisms:

1. **Experience-Following Property**: a high input-similarity between the current task and a retrieved memory record yields a high output-similarity (Pearson r ≈ 1.0). Agents *imitate* retrieved demonstrations — this is how memory helps AND how it harms. A retrieved noisy/misaligned record is imitated and amplifies the error.
2. **Error Propagation**: if a retrieved memory record contains noisy/incorrect outputs, the agent replicates and amplifies the error; if that execution is re-added to memory, the error propagates. This is the memory analogue of the BenchJack evaluator-gaming finding.
3. **Misaligned Experience Replay**: some records that pass the evaluator's quality filter still consistently lead to poor downstream execution because they are misaligned with the current task distribution.

Quantitatively, **add-all is worse than fixed-memory** on 3 of 4 agents tested; **strict-evaluator + selective addition beats add-all by 22–25 points**; and **history-based deletion** (remove a record retrieved ≥n times if its average downstream utility is below threshold) *improves* performance beyond no-deletion on real agents. A key result: a small fine-tuned quality evaluator (300 training examples) recovers most of an oracle evaluator's gain, beating a large generic LLM judge — "directly applying a vanilla LLM as a trajectory evaluator may lead to a more severe negative impact than manually crafting a small but high-quality dataset."

This is the mechanism behind the SkillHone "repeating a previously successful repair that no longer matches" failure (2026-07-24): the agent retrieves a similar-looking past decision and imitates it, even if the current failure mode differs.

### Proposed Solution
Replace add-all with a selective memory policy, applied to the pipeline's own artifacts (research reports, issues, implementation logs) and as a recommendation for the agent's memory/skills:

1. **Selective-addition gate**: only store an artifact if a quality evaluator judges it high-quality. Start with a small specialized quality classifier (the paper shows 300 training examples suffice) rather than a generic GPT-judge-everything — the generic LLM judge can be net-negative.
[...]

PR #1455, opened 2026-07-28
feat: capture the agent's real tool calls for the evolution pipeline (Closes #1363)

### LINKED ISSUE #1363: [BLOCKER] No production capture of real agent tool calls — three issues and six PRs have failed on this one missing input
## The pattern

Three issues are stuck in the same loop, each with two closed PRs — six in total, all closed by the owner as *dead code* or *incoherent*:

| Issue | PRs closed | Owner's verdict on the wired attempt |
|---|---|---|
| #1267 adversarial evaluator floor test | #1274, #1279 | "does not execute real evaluators… derives synthetic scores from metadata, reports deterministic false failures" |
| #1268 tool-use competency rubric | #1276, #1281 | "reads the funnel's own synthetic trajectory record rather than real agent tool calls… assigns every call turn 0" |
| #1270 selective memory addition gate | #1275, #1280 | "runs after artifacts have already been written… its historical inputs are permanently empty" |

Different features, same failure: each needs a record of **what the agent actually did**, and no such record exists. Every implementation therefore invents a substitute, and every substitute is correctly rejected.

## Verified root cause

Two trajectory mechanisms exist on `main`. Neither produces the needed data:

1. **`scripts/evolution_trajectory_logger.py`** — the only production caller is `scripts/evolution_funnel.py`, and `tests/scripts/test_evolution_trajectory_wiring.py` (#1215) documents exactly what it records: an entry with `tool='evolution_funnel'`. The pipeline logs *its own* invocation. No `agent/`, `tools/`, `run_agent`, or `hermes_cli` module calls it — verified by grep across `main`.

2. **`agent.save_trajectories`** (`agent/agent_init.py:291`) — writes real conversation JSONL, but defaults to `False` and nothing in the evolution pipeline reads its output.

So `evolution/trajectories/` contains one synthetic self-record per funnel run. Any rubric computed over it measures the pipeline, not the agent.

## Why this is worth fixing first

It is not three feature problems, it is one input problem. Until real tool-call capture exists, each of these issues can only be implemented against synthetic data — which is why re-attempting them without this produces another incoherent PR. Six closed PRs is the measured cost of not having it.

## Proposed shape

1. Decide the source of truth: turn on `save_trajectories` for evolution-relevant runs, or add an explicit tool-call sink in the agent loop.
2. Define the record: `(session_id, turn_index, tool, args_hash, outcome, error_class, latency)` — the fields #1268's rubric needs for discovery / parameterization / error-recovery / efficiency, and enough for #1267 and #1270.
3. Write to a stable location the evolution stages already read, with retention.
4. Only then unblock #1267 / #1268 / #1270 against real data.

## Relationship

Blocks #1267, #1268, #1270 — all three carry the `blocked` label today with no issue naming what blocks them. This is that issue.

Lexus2016:
Closes #1363 — the blocker seven issues wait on.

## The gap

The pipeline had **no record of what the agent actually did**:

- `evolution_funnel` writes a trajectory containing one entry — `tool='evolution_funnel'`. The pipeline logging its own invocation.
- `agent.save_trajectories` writes real conversation JSONL, but defaults off, lands in the process CWD, and nothing in the pipeline reads it.
- `~/.hermes/evolution/trajectories/` — **0 files**.

That one missing input has cost **six PRs**. #1267, #1268 and #1270 each had two closed by the owner as *dead code* or *incoherent*, every time with the same verdict: the implementation *"reads the funnel's own synthetic trajectory record rather than real agent tool calls"*. Each attempt had to invent a substitute for data that did not exist.

Two of the waiting issues (#1359, #1436) state in their bodies that these trajectories already exist. #1359 literally parses the empty directory.

## What this adds

**The conversion, not new storage.** `agent/tool_call_capture.py` turns a finished turn's messages into the `TrajectoryLog` that `evolution_trajectory_logger` already defines and already knows how to persist. Wired in `finalize_turn` beside the existing `_save_trajectory` call, guarded like every other step there (#8049) — a metrics write must never discard a completed turn's response.

## Privacy — why this can run while `save_trajectories` stays off

That flag writes the **full ShareGPT conversation, user prose included**, which is exactly why it defaults off, and why turning it on for evolution would be the wrong trade.

This captures **call metadata only**: tool name, arguments through the logger's existing `redact_args`, a status, a truncated result summary. No user message, no assistant prose. `test_no_user_prose_reaches_disk` asserts it.

## The consumers want different shapes, so the record carries the union

| Issue | Needs | Field |
|---|---|---|
| #1268 tool-use rubric | per-call tool / args / outcome / order | `entries` |
| #1359 ERL heuristics | task-level outcome — is this trajectory worth distilling from? | `completed` |
| #1436 EMG graph matching | **pair** a failed and a successful run of the same task | `task_key` |

`task_key` is a 16-char hash of the task descriptor, not the text: pairing only needs equality, and the descriptor is user prose that must not enter the store.

Both new fields are **optional** on `TrajectoryLog` and omitted from `to_dict()` when unset, so the existing cron-stage caller and any reader written against the pre-#1363 shape are unaffected. `load_trajectory` reads them back, with absent meaning *"not recorded"* rather than *"failed"* — a distinction consumers must not collapse.

## Two deliberate extraction choices

- **A call with no matching result is kept as `pending`, not dropped.** A call that never returned is itself signal for #1268's error-recovery dimension.
- **Status is classified by importing `introspection_extract._tool_result_failed`**, not reimplemented. Two disagreeing definitions of failure is how a metric drifts from what it claims to measure.

## Tests

25 covering extraction, malformed input (8 shapes, none raise), redaction, pairing, back-compat of the old on-disk shape, and that capture never raises on an unwritable path. Plus 18 passing in the existing trajectory logger/store suites. Ruff clean.

## What this does not do

It does not enable anything by default beyond the capture itself, and it does not implement any of the seven waiting issues — it gives them a real input to be implemented against. Whether the record satisfies each of them should be checked when they are picked up, starting with #1268 (the flat-log consumer) and #1436 (the most demanding, needing pairs).

### DIFF
--- a/agent/tool_call_capture.py
+++ b/agent/tool_call_capture.py
@@ -19,13 +19,19 @@ turn's message list into the ``TrajectoryLog`` that ``evolution_trajectory_logge
 already defines and already knows how to persist under
 ``~/.hermes/evolution/trajectories/``.
 
-Privacy — why this can be on when ``save_trajectories`` is off
---------------------------------------------------------------
+Opt-in, and narrower than ``save_trajectories``
+-----------------------------------------------
 ``save_trajectories`` writes the full ShareGPT conversation, user prose
-included, which is why it is off by default and why turning it on for evolution
-would be the wrong trade. This captures **call metadata only**: tool name,
-redacted arguments (via the logger's existing ``redact_args``), a status, and a
-short result summary. No user message, no assistant prose, no file contents.
+included. This captures **call metadata only**: tool name, redacted arguments
+(via the logger's existing ``redact_args``), a status, and a short result
+summary. No user message, no assistant prose, no file contents.
+
+It is still **off by default**, behind ``HERMES_EVOLUTION_CAPTURE``. Redaction
+catches credential-shaped *keys*, not sensitive *values* in ordinary fields —
+a path with a username, an SQL string, an internal hostname all survive it — so
+writing on every turn of every interactive session unasked would be the wrong
+default even for metadata. The evolution cron stages enable it for their own
+runs, which is the behaviour the pipeline actually needs to measure.
 
 The consumers want different shapes, so the record carries the union
 -------------------------------------------------------------------
@@ -43,7 +49,10 @@ on, opaque enough to store.
 from __future__ import annotations
 
 import hashlib
+import hmac
 import json
+import os
+import secrets
 import sys
 from pathlib import Path
 from typing import Any, Dict, List, Optional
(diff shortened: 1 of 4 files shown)


Lexus2016:
Three defects found by cross-provider review (Gemini via consilium), all reproduced before fixing, all now in this PR (`b37d6277c`).

## 1. Data loss — multi-turn sessions kept only their last turn

`TrajectoryLog.save()` writes `<date>_<session>.json`, so **every turn overwrote the previous one**. Reproduced before the fix:

```
three turns of session s1 -> 1 file
  2026-07-28_s1.json: tools=['terminal']
  => first two turns LOST
```

The action *sequence* is precisely what the consumers of #1363 want, so losing it defeats the feature. Added `TrajectoryLog.append()` — one JSON object per line to `<date>_<session>.jsonl`. `save()` is untouched: overwrite is correct for the cron stage, which writes one trajectory per stage run per day.

After:
```
2026-07-28_s1.jsonl: 3 ходів -> ['read_file', 'patch', 'terminal']
```

## 2. Wrong default — capture ran unconditionally

I had argued redacted metadata was safe enough to run always. That was wrong: `redact_args` catches credential-shaped **keys**, not sensitive **values** in ordinary fields. A path containing a username, an SQL string, an internal hostname all survive it — and it grows unbounded.

Now behind `HERMES_EVOLUTION_CAPTURE`, **default OFF**, checked inside `capture_turn` so no caller can forget it. The evolution cron stages enable it for their own runs, which is the behaviour the pipeline actually needs to measure.

## 3. Reversible pairing key

`task_key` was `sha256(descriptor)[:16]` — deterministic and unsalted, so a short templated prompt (*"run the tests"*, *"fix issue #102"*) is recoverable from a rainbow table by anyone who can read the store. That defeats the point of hashing it.

Now an HMAC under a host-local salt kept beside the trajectories it keys, minted `0600` on first use:

```
store A: f70300a2aca1337c
store B: 373492eec77c8b1e     same text, different stores -> different keys
stable within a store: True   pairing still works, which is all #1436 needs
```

## Confirmed correct

The **seam**. `finalize_turn` is where calls and results are both settled and the turn-level outcome is known, and it is where the existing trajectory save already lives. Per-call middleware would have needed a plugin registration and manual state stitching for strictly less context.

Collision risk on a 16-hex key was also checked and dismissed: 64 bits needs ~4.3 billion distinct descriptors for a 50% collision.

34 tests now (9 new covering opt-in, append, and salting), 52 across this and the existing trajectory suites; ruff clean.

PR #1459, opened 2026-07-28
fix: tool_search streak counter fires injection at 3 and tool_describe at 5 (Closes #1373)

### LINKED ISSUE #1373: [FIX] tool_search streak counter not firing — 17 sessions identical to pre-merge baseline
The tool_search streak counter and fallback directive (merged in #1153) is not reducing the pattern. tool_search still shows 17 sessions with max 9 consecutive searches — identical to the pre-merge baseline. The agent reformulates search queries up to 9 consecutive times without issuing a tool_call, indicating the streak counter is either not firing or the fallback nudge is not changing behavior.

- Frequency: 17 sessions with max 9 consecutive tool_search calls (7-day window)
- Tool/area involved: tool_search
- Failure shape: repeated search query reformulation without converging on a tool to invoke (tool_call)
- Prior fix: #1153 added per-session streak counter + fallback_directive after N consecutive searches with no tool_call (merged 2026-07-18)
- Post-merge verdict: no-signal (2026-07-23: "tool_search still at 16 sessions with max 9 consecutive in latest 7d extract — identical to pre-merge baseline")

### Impact on real tasks
The agent burns turns searching for tools instead of using them. 9 consecutive searches without a tool_call means the agent cannot find the right tool and keeps reformulating its query. This blocks tasks that require deferred-tool invocation, particularly when the agent needs an MCP or plugin tool it hasn't used before.

### Proposed direction
The fallback directive after N searches is not changing behavior. The fix should:
1. After 3 consecutive tool_search calls with no tool_call, inject the FULL list of available tool names (not just a nudge) so the agent can see all options at once.
2. Track and surface the previous search queries in the directive so the agent can see what it already tried and avoid repeating.
3. Consider auto-invoking tool_describe on the top result after 5 consecutive searches, forcing the agent to at least examine one tool rather than continuing to search.

### Value
- Impact: 0.5 (17 sessions, blocks deferred-tool discovery)
- Effort: 0.3 (inject tool list + auto-describe after threshold)
- Priority Score: 0.5 × 2 × (1 − 0.4 × 0.3) = 0.88

Lexus2016:
Automated evolution PR for issue #1373.

Root cause: callers pass session_id=agent.session_id or empty string, and the old streak counter treated any falsy value (incl. empty string) as do-not-track, so the streak never incremented and the fallback directive never fired.

Fix:
- Map empty-string session_id to stable key __default_session__ so production fires; None still opts out (preserves pure-function test contract).
- Add per-session previous-queries tracking (rolling, capped at 8).
- At streak >= 3 (search_streak_threshold): inject full tool list + previous queries + fallback directive.
- At streak >= 5 (new search_streak_describe_threshold): additionally inject auto_describe with the top hit full schema.

Tests: 8 new tests in TestSearchStreakEmptySessionNowFires covering all four spec requirements.

### DIFF
--- a/tests/tools/test_tool_search.py
+++ b/tests/tools/test_tool_search.py
@@ -20,7 +20,9 @@ if _REPO_ROOT not in sys.path:
     sys.path.insert(0, _REPO_ROOT)
 
 
-def _td(name: str, description: str = "", properties: Dict[str, Any] | None = None) -> Dict[str, Any]:
+def _td(
+    name: str, description: str = "", properties: Dict[str, Any] | None = None
+) -> Dict[str, Any]:
     return {
         "type": "function",
         "function": {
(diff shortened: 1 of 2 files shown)


Lexus2016:
Reviewed and merging. This correctly implements the root cause my diagnostic PR (#1445) surfaced, and goes further by adding `previous_queries` so the model can see it is circling.

One risk worth recording rather than blocking on: **all empty-session callers now share a single `__default_session__` counter.** Sub-agents built by `delegate_task` run in the same process, so if a parent and a child both reach the counter with an empty session id, their streaks add together and the directive fires earlier than either warrants.

Worst case is a nudge one or two searches early, which is mild — and the alternative (no counter at all) is what this PR is fixing. Worth watching in the next digest: if `tool_search` streak counts drop but the directive appears in sessions that only searched twice, this is why.

Note the diagnostic warning from #1445 is removed here, which is right — there is no longer a silent-off path to warn about. If a future change reintroduces one, the lesson is that a counter returning 0 forever looks identical to a counter that is working.

Issue #1469, opened 2026-07-28
[FEATURE] OATS: offline embedding refinement for tool-selection accuracy (zero serving-time cost)


Lexus2016:
## Context

**Source:** arXiv:2603.13426 — "Outcome-Aware Tool Selection for Semantic Routers" (Chen et al., Mar 2026)
- https://arxiv.org/abs/2603.13426

OATS improves tool-selection accuracy by **interpolating tool embeddings toward the centroid of queries where each tool historically succeeded** — run as a batch job, zero cost at serving time. On MetaTool benchmark, NDCG@5 improved from 0.869 to **0.940**; on ToolBench, 0.834 to **0.848**.

The technique is simple: after each cycle, collect `{query_embedding, tool_called, outcome}`, compute per-tool success-centroids, and nudge each tool's description embedding toward its success centroid. At inference time, the refined embeddings are already baked in — no runtime overhead.

## Why it matters for Hermes

The skill-discovery eval (07-26) already identified that skill/tool description quality drives selection accuracy, but the proposed remediation was *rewriting descriptions* (a human task). OATS is the **automated counterpart**: instead of rewriting descriptions, *learn* the embedding shift from observed outcomes. This maps directly onto Hermes's cron architecture — a nightly job that refines tool/skill embeddings from the day's `{task → tool_used → outcome}` telemetry.

## Tasks

- [ ] Collect `{query_embedding, tool_called, outcome}` tuples per cycle (data the pipeline already gathers)
- [ ] Compute per-tool success-centroids: average query embedding for calls where the tool succeeded
- [ ] Interpolate each tool's description embedding toward its success centroid (interpolation rate as hyperparameter)
- [ ] Store refined embeddings in the embedding store (replacing or augmenting baseline embeddings)
- [ ] Add a nightly cron job that runs the refinement batch
- [ ] Benchmark: measure tool-selection accuracy before/after on the skill-discovery gold-set

## Acceptance criteria

- Tool-selection accuracy improves on the eval gold-set after refinement
- Zero runtime overhead at serving time (refinement is batch-only)
- Nightly cron runs successfully and updates embeddings

## Research reference

Research report 2026-07-29, New Features §3. Priority score: 1.12.

Lexus2016:
Duplicate — this idea was already filed and already judged.

[...]

Issue #1460, opened 2026-07-28
[FEATURE] MCP 2026-07-28 stateless spec migration — audit and update all MCP integrations


Lexus2016:
## Context

The MCP spec finalized its largest-ever revision on **2026-07-28**. The core is now **stateless**: the `initialize`/`initialized` handshake and `Mcp-Session-Id` header are **removed**, replaced by per-request protocol-version metadata and a new `server/discover` RPC. Deprecated features remain functional for 12 months but backward compatibility with older clients is **NOT guaranteed**.

**Sources:**
- https://blog.modelcontextprotocol.io/posts/2026-07-28-release-candidate
- https://www.theregister.com/devops/2026/07/23/model-context-protocol-prepares-to-break-with-its-stateful-past/5276722

## Why it matters

Stateless core enables horizontal scaling behind load balancers without sticky sessions. But any MCP server/client integration that depends on the session handshake will break. This is the single most time-sensitive item — the 12-month deprecation clock starts now.

## Tasks

- [ ] Audit all MCP server and client integrations for stateful assumptions (`initialize` handshake, `Mcp-Session-Id` header, session-scoped state)
- [ ] Update MCP SDK dependencies to the latest version supporting the 2026-07-28 spec
- [ ] Implement per-request protocol-version metadata forwarding
- [ ] Add `server/discover` RPC support as a replacement for the discovery handshake
- [ ] Implement backward-compatibility fallback for older clients (12-month window)
- [ ] Add integration tests that verify stateless operation behind a load balancer (no sticky sessions)
- [ ] Document migration path for any custom MCP servers in the skills/plugins ecosystem

## Acceptance criteria

- All MCP integrations operate correctly without session state
- Tests pass with requests routed to different server instances (no sticky sessions)
- Legacy clients still work during the deprecation window

## Research reference

Research report 2026-07-28, Finding #1. Priority score: time-critical (breaking change, deprecation clock active).

Lexus2016:
Consolidated into #1287, which tracks this migration and already has Slice A (the stateful-surface audit) landed.

Not a pure duplicate — this issue carries genuinely new information the parent did not have: the spec finalized **today (2026-07-28)**, there is a new `server/discover` RPC absent from the parent's audit, and deprecated features remain functional for **12 months**. That last point is what lifts the hold: the parent deferred Slice B because a stateless rewrite would break every 2025-spec server, and a 12-month deprecation window changes that calculus.

[...]

Issue #1437, opened 2026-07-28
[FEATURE] AgentSentry: counterfactual re-execution at tool-return boundaries for indirect prompt injection detection


Lexus2016:
## Summary

AgentSentry detects indirect prompt injection by running **counterfactual re-executions** at tool-return boundaries: for each untrusted tool output that enters the agent's context, it launches ~4 shadow variants with different portions of the untrusted content masked, then estimates causal dominance — if the agent's next action changes materially depending on *which part* of the tool output is visible, that output is steering the agent, not informing it. On AgentDojo, AgentSentry eliminates successful attacks (74.55% Utility Under Attack preserved — no successful injections), with low false-positive rate.

This is structurally different from Hermes's current injection defenses (system-prompt instructions to "ignore embedded commands", the research skill's security rules). Those are *advisory* — they tell the model what to ignore. AgentSentry is *empirical* — it detects injection by observing causal effect on behavior. The post-turn assertion hook (#1301, merged 07-27) checks *outcome* state; AgentSentry checks *input causality* — they operate at different layers. The ExploitGym incident (07-28 report) demonstrated that sandbox escape is real and autonomous; a causal-dominance check at tool-return boundaries is the runtime defense layer that complements sandbox isolation.

## Source

- **Paper**: arXiv:2602.22724 — "AgentSentry: Mitigating Indirect Prompt Injection via Temporal Causal Diagnostics and Context Purification" (Zhang et al., Feb 26 2026)
- **Link**: https://arxiv.org/abs/2602.22724
- **Frontier standing**: behind
- **Related in-repo**: post-turn assertion hook (#1301), ExploitGym incident (07-28 report)

## Implementation Path for Hermes

The counterfactual re-executions are the expensive part (~4 extra forward passes per tool call with untrusted content); the paper targets only tool outputs flagged as untrusted, which keeps the overhead bounded.

1. **Untrusted-output flagging**: mark tool outputs that originate from external sources (web pages, emails, files read from disk, MCP server results) as untrusted at the tool-return boundary. Most tool outputs (local computation, file writes) are trusted.
2. **Counterfactual re-execution hook** (`scripts/evolution_agentsentry.py` or a runtime hook): for each untrusted tool output, launch ~4 shadow forward-pass variants with different masked portions of the untrusted content. Capture the agent's proposed next action for each variant.
[...]

Issue #1435, opened 2026-07-28
[IMPROVEMENT] pass^k reliability metric for the eval suite (stop measuring luck)


Lexus2016:
## Summary

**pass@k** measures "at least one of k attempts succeeds" — it rewards luck. **pass^k** measures "ALL k attempts succeed" — it rewards reliability. The gap can be **63 percentage points** on the same task (a task where the agent succeeds 1/5 of the time scores pass@5=1.0 but pass^5=0.0).

The evolution pipeline currently grades pass/fail per cycle; if it runs N attempts and reports "passed", that's pass@1 or pass@N — neither captures whether the agent can *reliably* reproduce the result. Combined with the paper's finding on **discriminative task selection** (selecting tasks that maximize score variance between models yields 5–10× benchmark cost reduction), this is a zero-cost methodology upgrade.

This is the metric-level complement to the 07-27 artifact-grading proposal (#1302, which changed *what* is measured); pass^k changes *how reliably* it's measured.

## Source

- **SE-Agent**: arXiv:2508.02085 (NeurIPS 2025) — https://arxiv.org/abs/2508.02085
- **HAL meta-evaluation** baseline (21,730 rollouts, 07-25 baseline)
- **Multiple agent-eval methodology papers**, Jul 2026
- **Frontier standing**: at-par (Hermes *can* run N attempts today — the improvement is making the metric explicit)

## Implementation Path for Hermes

1. **Report pass^k alongside pass@k** for the eval gold-set: run N attempts per task, compute both metrics. `pass^k = (all N succeeded) ? 1 : 0`; `pass@k = (≥1 of N succeeded) ? 1 : 0`.
2. **Add to the eval reporting** (`scripts/` eval harness or `metrics.jsonl`): store both per-task and aggregate pass^k / pass@k.
3. **Discriminative task selection**: prioritize tasks that have high pass^k variance (tasks where some agent versions reliably succeed and others don't) — this yields 5–10× benchmark cost reduction per the paper.
4. **No new infra needed** — Hermes can already run N attempts; this makes the metric explicit and uses it to filter the eval set.

## Impact / Effort

- **Impact**: Medium — the reliability signal (pass^k) changes which improvements are trusted; an improvement that raises pass@k but drops pass^k is fragile.
- **Effort**: Low — a metric computation over attempts already being run; zero new infrastructure.
- **Priority Score**: 1.12 (impact 0.6 × 2 × (1 − 0.4 × effort 0.1))

[...]

PR #1471, opened 2026-07-28
fix(evolution): stop re-filing rejected ideas — dedup must see closed issues (Closes #1470)

### LINKED ISSUE #1470: [FIX] Rejected ideas are re-filed every cycle — the duplicate check only looks at OPEN issues
## Problem

The pipeline re-files ideas it has already rejected, indefinitely. Nine confirmed cases in a single cycle today.

At **11:03 UTC** evolution-analysis rejected five proposals as out-of-scope:

| Issue | Verdict |
|---|---|
| #1433 | *"OATS offline embedding refinement — speculative, no concrete implementation path. Fails focus-test."* |
| #1434 | *"Detect Unfaithful Safety Refusal — speculative detection framework. Fails focus-test."* |
| #1435 | *"pass^k reliability metric — speculative eval proposal. Fails focus-test."* |
| #1437 | *"AgentSentry counterfactual re-execution — speculative multi-agent framework, heavy infra. Fails focus-test."* |
| #1343, #1345, #1313 | rejected earlier, same shape |

**Within the same cycle**, the research → issues stages filed all of them again: #1469 (=#1433), #1467 (=#1434), #1468 (=#1435), #1462 (=#1437), #1464 (=#1343), #1466 (=#1345), #1465 (=#1313). Plus #1461 (=#1436) and #1460 (=#1287), where the originals are still live.

Nine duplicates in one cycle. The backlog cannot converge: every rejection is undone by the next research pass.

## Root cause

`skills/evolution/evolution-issues/SKILL.md:187`, the duplicate check run immediately before `gh issue create`:

```bash
existing=$(gh issue list --repo "$REPO" --state open --search "in:title $TITLE" \
  --json number,title --jq ".[] | select(.title==\"$TITLE\") | .number" | head -1)
```

**`--state open`.** A rejected issue is closed, so it is invisible to the check, so the idea is filed again. The rejection is what makes it re-fileable.

Note line 85 of the same file already fetches `--state all` for the earlier survey step — the mechanism is understood, it just is not used at the point that matters.

## Compounding factor: wrong close reason

The rejections are recorded as `stateReason: COMPLETED`, not `not planned`. So even a check that did look at closed issues would read *"this was done"* rather than *"this was rejected"*, and could reasonably re-file on the assumption that the work has since regressed.

Same bookkeeping defect as #1272/#1273, which were closed COMPLETED after their PR was closed unmerged.

## Proposed fix

1. **Change the pre-create check to `--state all`.** A title match on a closed issue must block creation.
2. **Distinguish the close reason.** Blocked on `not planned` (rejected — do not re-file); allowed on `completed` only when the new proposal cites evidence the fix regressed, and then it should say so in the body.
3. **Reject with `--reason "not planned"`.** An idea judged out-of-scope was not completed.
4. **Record rejections in the dedup cache** the skill already maintains (#91), so the short-circuit happens before the API call rather than after.

## Why this matters more than any single proposal in the backlog

Every rejected idea returns, so analysis re-triages it, consuming a cycle to reach the verdict it already reached. Nine duplicates today is nine wasted triage decisions and nine issues a human has to read past. The rejection filter has no memory, which makes the focus-test unenforceable in practice.

---
*Filed while clearing the backlog, from the duplicate pattern it exposed.*

Lexus2016:
The pipeline re-files ideas it has already rejected, indefinitely. **Nine duplicates landed in a single cycle today.**

At **11:03 UTC** analysis rejected #1433, #1434, #1435, #1437 (and earlier #1313, #1343, #1345) as out-of-scope — *"speculative, no concrete implementation path. Fails focus-test."* The same cycle's research/issues pass filed **every one of them again** as #1460–#1469.

## Root cause

`evolution-issues/SKILL.md:187`, the check run immediately before `gh issue create`:

```bash
gh issue list --state open --search "in:title $TITLE"
```

**`--state open`.** A rejected idea is a *closed* issue, so the check cannot see it, so it is filed again. The rejection is precisely what makes it re-fileable.

The backlog therefore cannot converge: every rejection is undone by the next research pass, and analysis spends another triage reaching the verdict it already reached. Nine duplicates is nine wasted triage decisions and nine issues a human reads past.

## Two changes, one per stage

**evolution-issues** — `--state all`, and read `stateReason` to decide what a hit means:

| State | Meaning |
|---|---|
| `NOT_PLANNED` | rejected → do **not** re-file; the focus-test verdict stands |
| `COMPLETED` | shipped → new issue only with **evidence it regressed**, cited in the body |
| `OPEN` | already tracked → skip |

Line 85 of the same file already fetched `--state all` for the survey step, so the mechanism was understood — it just was not used at the point that decides.

**evolution-analysis** — close rejections with `--reason "not planned"`.

The default `completed` tells every later reader the work was **done**, and that is exactly the field the duplicate check above reads. A rejection recorded as completed is indistinguishable from a shipped fix, so it gets re-filed as a suspected regression. The `rejected` label does not carry this — the state reason does.

Same bookkeeping defect as #1272/#1273, which were closed `COMPLETED` after their PR was closed unmerged.

## Why this one first

It is the reason the backlog does not shrink. Every other proposal in the queue competes for triage attention with ideas that were already turned down and came back.

22 passing in `test_evolution_skill_integrity.py`, which lints every skill reference and command against the real repo.

### DIFF
--- a/skills/evolution/evolution-analysis/SKILL.md
+++ b/skills/evolution/evolution-analysis/SKILL.md
@@ -167,12 +167,21 @@ gh issue view <N> --repo Lexus2016/hermes-agent-evolution --json body,comments
 
    CLOSE every rejected issue with a clear reason + the canonical `rejected`
    status label, so the backlog shows at a glance what was turned down (the
-   *why* is in the closing comment) and the same idea isn't re-proposed:
+   *why* is in the closing comment) and the same idea isn't re-proposed.
+
+   **`--reason "not planned"` is required, not cosmetic (#1470).** Closing a
+   rejection as the default `completed` tells every later reader the work was
+   DONE. The issues stage's duplicate check reads exactly that field to decide
+   whether an idea may be re-filed: `NOT_PLANNED` blocks it, `COMPLETED` allows
+   it as a possible regression. Rejections closed as `completed` are therefore
+   re-filed on the next research pass — nine landed in one cycle that way
+   (#1460-#1469). The label alone does not carry this; the state reason does.
    ```bash
    # Ensure the status label exists (idempotent), then close + label:
    gh label create rejected --color b60205 \
      --description "Not accepted by evolution — see closing comment" 2>/dev/null || true
    gh issue close <N> --repo Lexus2016/hermes-agent-evolution \
+     --reason "not planned" \
      --comment "Rejected by evolution-analysis: <already-exists|out-of-scope|harmful|duplicate> — <one-line reason>."
    gh issue edit <N> --repo Lexus2016/hermes-agent-evolution \
      --add-label rejected --remove-label needs-work 2>/dev/null || true
--- a/skills/evolution/evolution-issues/SKILL.md
+++ b/skills/evolution/evolution-issues/SKILL.md
@@ -180,13 +180,22 @@ Then create:
    local cache check above does NOT protect against a `gh issue create` that
    silently succeeds but whose response times out, so you "retry" and file a true
    duplicate (this happened — #193/#194, identical body, 13s apart, one proposal).
-   So immediately before creating, confirm no OPEN issue already has this exact
-   title; if one does, SKIP the create and just record it:
+   So immediately before creating, confirm no issue already has this exact
+   title — **including closed ones**; if one does, SKIP the create and just
+   record it:
 ```bash
 TITLE="[FEATURE] <short title>"   # right prefix per category: [FEATURE]/[FIX]/[IMPROVEMENT]/...
-existing=$(gh issue list --repo "$REPO" --state open --search "in:title $TITLE" \
-  --json number,title --jq ".[] | select(.title==\"$TITLE\") | .number" | head -1)
+existing=$(gh issue list --repo "$REPO" --state all --search "in:title $TITLE" \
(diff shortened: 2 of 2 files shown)


Lexus2016:
Considered extending `evolution_pre_submit_triage.py` (the deterministic gate at step 2c) to closed issues as well, and deliberately did **not**.

That gate documents its scope explicitly — *"Closed issues are out of scope for this gate (handled in SKILL.md), so they never produce a skip signal"* — and its whole design is an anti-fabrication guard: it skips only on a Dice ≥ 0.85 near-exact title overlap, and defaults to CREATE on any doubt, because the project's documented failure mode is triage **fabricating** a rejection and wrongly closing a real issue (#83/#101).

Widening it to closed issues would increase wrongful-skip risk against exactly that failure mode, and a wrongful skip silently suppresses a genuine proposal.

The fix in this PR sits at step 3 instead — the exact-title check immediately before `gh issue create`. That is safe to widen because it matches on **exact title equality**, not similarity, so it cannot fabricate a rejection: either the title is character-for-character identical to an existing issue, or it is not.

So the layering ends up: fuzzy gate on open only (conservative), exact-title check on all states (deterministic), and the by-meaning comparison in step 2b unchanged.

Issue #1454, opened 2026-07-28
[FIX] Security-scan approval-block fallback directive (Child 3 of #1371)


Lexus2016:
## Parent: #1371 (terminal 27.4% failure rate — dominant signal)

Slice 3 of the proposed 3-way split. Standalone — a distinct failure class (security-scan blocks) with its own code path.

## Scope (~50 lines)

1. When a terminal command is blocked by the security-scan approval flow, surface a **fallback directive**: the block reason + the alternative (request approval, or rephrase the command to avoid the trigger).
2. Today the block is reported but the agent often just retries the same command; the directive makes the alternative explicit.
3. Dovetails with #1439 (runtime circuit-breaker for repeated-identical tool-failure spirals) — this slice handles the *security-scan* sub-class, #1439 the general cross-tool class.

## Acceptance criteria
- [ ] Security-scan block surfaces the block reason + fallback directive
- [ ] Retrying the identical blocked command does not silently loop
- [ ] Unit test pins the directive

---
*Child 3 of #1371. Subsumes the security-scan portion; the general cross-tool circuit-breaker lives in #1439.*

Lexus2016:
Already implemented — closing rather than re-building it.

Both paths this issue asks for exist in `tools/terminal_tool.py` today.

**Non-interactive block** (cron/subagent, no one to approve) — line ~2583:
```
BLOCKED: Command requires approval ({desc}) but no interactive user or gateway
is present to approve it in this {cron|subagent} context. Do NOT retry this
command — find an alternative approach using file/search tools instead of
terminal, or set approvals.cron_mode: approve in config.yaml if this command is
safe to auto-approve.
```

**Interactive denial** — line ~2615:
```
Command denied: {desc}. Use the approval prompt to allow it, or rephrase the
command.
```

Against this issue's own scope:
- *"the block reason"* → `{desc}` in both, plus `pattern_key` and `description` on the non-interactive envelope
- *"request approval"* → covered by both messages
- *"rephrase the command to avoid the trigger"* → explicit in the interactive one
- *"the agent often just retries the same command"* → the non-interactive path says **"Do NOT retry this command"** verbatim and names the alternative tools

Covered by `tests/tools/test_approval.py`.

## What was actually missing in this failure class

[...]

Issue #1453, opened 2026-07-28
[FIX] Strategy-switch directive after 3 consecutive terminal failures (Child 2 of #1371)


Lexus2016:
## Parent: #1371 (terminal 27.4% failure rate — dominant signal)

Slice 2 of the proposed 3-way split. **Depends on:** Child 1 (#1452) — the enriched errors make the strategy-switch directive actionable.

## Scope (~70 lines)

1. After **3 consecutive** terminal failures (any error class, not just identical — generalizes #1022 which caps only identical-failure spirals), inject a system-level directive nudging the agent to switch strategy: re-read the task, try a different tool (e.g. `patch` instead of `terminal sed`), or ask for clarification.
2. The directive is a short prompt-level nudge appended to the next turn context, not a hard block — the agent can still proceed if it has a reason.
3. Resets on the first success after the run of failures.

## Why second
The introspection data shows up to 55 consecutive retries in a single session. Existing #1022 handles *identical* retries; this catches *varied-but-failing* retries where the agent flails across different commands without changing approach.

## Acceptance criteria
- [ ] 3 consecutive terminal failures inject the strategy-switch directive on the next turn
- [ ] A success resets the consecutive-failure counter
- [ ] The directive is advisory (no hard block)
- [ ] Unit test pins the counter + directive injection

---
*Child 2 of #1371.*

Issue #1371, opened 2026-07-27
[FIX] terminal 27.4% failure rate across 673 sessions — dominant signal drowning all others


Lexus2016:
## Problem (observed in real usage)
The terminal tool has a 27.4% failure rate — the highest of any tool by rate — with 2,020 failures across 673 sessions in a 7-day window. This dwarfs every other signal in the system. The agent repeatedly runs commands that fail (non-zero exit, timeout, or security-scan block) and retries, sometimes up to 55 consecutive times in a single session. Multiple prior fixes (spiral caps, circuit breakers, failure classifiers) have been merged and all regressed — the terminal failure count has grown monotonically from 174 → 1,237 → 1,574 → 2,020 across cycles.

### Evidence (aggregated, anonymized)
- Frequency: 2,020 terminal failures across 673 of 7,383 recent sessions (7-day window)
- Tool/area involved: terminal
- Failure rate: 27.4%
- Failure shape: non-zero exit codes, timeouts (141 total), and security-scan approval blocks; agent retries same or similar command up to 55 consecutive times
- Trend: 174 → 1,237 → 1,574 → 2,020 across successive 7-day windows — every fix cycle (including #974, #1167, #1022, #1091, #1109) was merged then regressed

### Impact on real tasks
Terminal is the agent's primary execution tool. A 27.4% failure rate means the agent fails on more than 1 in 4 commands it runs, and when it fails it retries extensively. This blocks builds, tests, git operations, installs, and any task requiring shell execution. The 55-deep spirals consume entire sessions without progress.

### Proposed direction
Prior fixes added spiral caps and classifiers but the failure rate itself has not dropped. The problem is that caps bound depth but don't prevent the failure:
1. Security-scan blocks (approval_pending in cron/subagent contexts) are a major failure class — the agent hits a block, cannot get approval, and retries. The fallback should be a non-terminal alternative (use read_file/write_file/search_files instead of terminal) injected directly into the error.
2. Non-zero exits often have empty or unhelpful error output (#898 fixed empty-error-on-511 but many other exit codes still produce opaque output). Enrich ALL non-zero exit errors with the actual stderr and a suggested next action.
3. After 3 consecutive terminal failures in a session, inject a strategy-switch directive: "You have failed 3 terminal commands — consider using read_file/write_file/search_files for filesystem operations instead."

### Value
- Impact: 1.0 (dominant signal, blocks all shell-based work)
[...]

PR #1472, opened 2026-07-28
fix(terminal): name shell-level exit codes 126/127 (Closes #1452)

### LINKED ISSUE #1452: [FIX] Enrich non-zero terminal exits with stderr + suggested action (Child 1 of #1371)
## Parent: #1371 (terminal 27.4% failure rate — dominant signal)

Slice 1 of the proposed 3-way split (from the #1371 comment). Standalone — touches only the non-zero-exit error path in `terminal_tool.py`.

## Scope (~60 lines)

1. When a terminal command exits non-zero, the error surfaced to the agent includes the **stderr tail** (already captured) AND a one-line **suggested action** derived from the exit code / error class (e.g. `127` → "command not found, check spelling or PATH"; `1` → generic failure, see stderr).
2. The suggested action is a small static lookup table keyed by exit code + a couple of regex heuristics on stderr (`permission denied` → check perms; `not found` → check install/PATH; `timed out` → raise timeout or simplify).
3. No behaviour change on success or timeout-141 beyond the existing message.

## Why first
Smallest, highest-confidence slice. Directly attacks the root cause (agent retries the *same* failing command because the error gives no direction). Existing spiral-detection (#1022) caps *identical* retries but does not enrich the error — enrichment reduces the retries at source.

## Acceptance criteria
- [ ] Non-zero exit error message contains the stderr tail + a suggested-action line
- [ ] The suggestion table has ≥5 entries covering the most common exit codes/classes
- [ ] Timeout (141) and success paths unchanged
- [ ] Unit test pins the enrichment for ≥3 exit codes

---
*Child 1 of #1371.*

Lexus2016:
Child 1 of #1371 (terminal 27.4% failure rate across 673 sessions — the dominant failure signal).

## The gap

`_interpret_exit_code` already explains **per-command** exit codes (`grep` 1 = no matches, `curl` 7 = connection failed) and **signal** codes (137 = SIGKILL). It said nothing about **126 and 127**, because those are not per-command — they are about the *invocation*. 127 means the shell never found anything to run, whatever the command was, so it falls straight through the per-command table and produced no note at all.

## Why these two specifically

The failure is not in the command's logic. **Re-running the same string cannot succeed** — which is exactly the retry-spiral shape behind the parent issue's numbers. So each note states the cause *and* what to change:

| Code | Note |
|---|---|
| **126** | found but not executable; check permissions (`chmod +x`) or whether it is a directory. Re-running unchanged will fail identically. |
| **127** | not found; check spelling, or whether the tool is installed and on PATH. Re-running unchanged will fail identically; try `which` or an absolute path. |

## Ordering is the risk, and it is tested

The new lookup is a **fallback**, matched after the per-command table and the signal codes. Tests pin this:

- `grep x f` exit 1 still reads *"No matches found (not an error)"*, not a generic shell hint
- `curl` exit 7 still reads *"Failed to connect to host"*
- `sleep` exit 137 still reads *"SIGKILL"*
- an unknown code still returns `None` rather than a fabricated explanation

8 new tests, 39 passing in `test_terminal_exit_semantics.py`, 345 across the terminal suites; ruff clean.

## Scope

Deliberately just this. Children 2 (#1453, strategy-switch after 3 consecutive failures) and 3 (#1454, security-scan block directive) are separate — and #1453 in particular needs checking against `loop_guard`, which already fires on the third consecutive terminal failure today.

### DIFF
--- a/tests/tools/test_terminal_exit_semantics.py
+++ b/tests/tools/test_terminal_exit_semantics.py
@@ -150,3 +150,50 @@ class TestInterpretExitCode:
     def test_only_env_vars(self):
         """Command with only env var assignments, no actual command."""
         assert _interpret_exit_code("FOO=bar", 1) is None
+
+
+class TestShellLevelExitCodes:
+    """126/127 are about the INVOCATION, not the program (#1452).
+
+    They are command-agnostic — 127 means the shell found nothing to run,
+    whatever the command was — so they fell through the per-command table and
+    produced no note at all. Worth naming because re-running unchanged cannot
+    succeed, which is the retry-spiral shape behind terminal's 27.4% failure
+    rate (#1371).
+    """
+
+    def test_127_names_the_cause_and_the_fix(self):
+        note = _interpret_exit_code("nosuchtool --flag", 127)
+        assert note is not None
+        assert "not found" in note.lower()
+        assert "path" in note.lower()
+
+    def test_126_distinguishes_permission_from_missing(self):
+        note = _interpret_exit_code("./script.sh", 126)
+        assert note is not None
+        assert "not executable" in note.lower()
+        assert "not found" not in note.lower()
+
+    def test_both_say_a_bare_retry_will_not_help(self):
+        """The point is to stop the identical re-issue, so it has to be said."""
+        for code in (126, 127):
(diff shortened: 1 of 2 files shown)


Lexus2016:

PR #1525, opened 2026-07-31
feat: disambiguate read_file/read_file_raw error classes (Closes #1488)

### LINKED ISSUE #1488: [FIX] read_file permission-denied + unclassified errors — 102 failures/7d, distinct from covered file-not-found bucket
read_file fails 102 times for reasons OTHER than file-not-found: 60 "other"
(encoding errors, symlinks, special files, truncated content) and 42
"permission-denied". All existing read_file issues (#1370/#886/#970) target the
file-not-found path-resolution class. The permission-denied and "other" buckets
are a DIFFERENT failure class that a path-resolution fix cannot address.

read_file is the highest-rate failing core tool (10.5% failure rate), so the
102 non-path-resolution failures are a major uncovered slice of the dominant
tool-failure signal.

- Frequency: 60 "other" + 42 "permission-denied" = 102 events (7-day window, 7616 sessions)
- read_file repeated in 223 sessions (max 20 consecutive)
- Covered bucket for comparison: 701 file-not-found (addressed by #1370/#886/#970)
- Failure shape: tool returns a generic error; agent cannot tell permission denial
  from encoding failure and retries the same path.

### Proposed direction
Disambiguate the error classes: distinct, actionable messages for
permission-denied (suggest checking permissions / running with elevation), encoding
errors (suggest a fallback decoder), and special files (suggest binary handling),
so the agent picks the right recovery instead of blind-retrying.

### Value
- Impact: 0.7
- Effort: 0.45
- Priority Score: 1.148

Lexus2016:
## Summary

Both `read_file` and `read_file_raw` used `wc -c < path 2>/dev/null` with stderr suppressed, then routed ALL non-zero exits to `_suggest_similar_files` — which returns a "File not found" error regardless of the actual cause. This meant permission-denied and directory errors were misclassified as `not_found`, causing the agent to blind-retry instead of fixing permissions or choosing a different tool.

**102 failures/7d** (42 permission-denied + 60 "other") traced to this single root cause. 223 sessions affected, max 20 consecutive retries.

## Fix

Route the failure through `_diagnose_read_failure` (already used by `patch_replace` since #1326), which probes with `test -e` / `test -d` to disambiguate:
- **Not found** → `_suggest_similar_files` (preserves old behavior for genuine misses)
- **Directory** → "Cannot read a directory: ..." (classifies distinctly from `not_found`)
- **Permission denied** → "Permission denied reading file: ..." (classifies as `permission`)

The method is generalized with an `operation` parameter (default "patch") so directory/permission messages are actionable for both patch and read callers.

## Files changed
- `tools/file_operations.py` — `read_file` and `read_file_raw` failure path + `_diagnose_read_failure` generalization
- `tests/tools/test_file_operations.py` — 5 new tests in `TestReadFileErrorDisambiguation`

## Verification
- `ruff check` ✓
- `pytest tests/tools/test_file_operations.py` — 102 passed ✓
- `pytest tests/tools/test_file_operations_edge_cases.py tests/tools/test_file_read_guards.py tests/tools/test_read_file_failure_rate.py` — 89 passed ✓
- Diff: 165 insertions + 20 deletions = 185 lines (within 200-line self-merge cap)

Closes #1488

### DIFF
--- a/tests/tools/test_file_operations.py
+++ b/tests/tools/test_file_operations.py
@@ -1001,3 +1001,137 @@ class TestPatchReplaceReadFailureClassification:
         assert "permission" in result.error.lower()
         # Distinct from the old generic "Failed to read file".
         assert "failed to read file" != result.error.strip().lower()
+
+
+# =========================================================================
+# #1488 — read_file / read_file_raw error disambiguation
+# =========================================================================
+# Before #1488, read_file and read_file_raw used ``wc -c < path 2>/dev/null``
+# with stderr suppressed, then routed ALL non-zero exits to
+# _suggest_similar_files — which returns a "File not found" error regardless
+# of the actual cause. This meant permission-denied and directory errors were
+# misclassified as not_found, causing the agent to blind-retry instead of
+# fixing permissions or choosing a different tool. 102 failures/7d (42
+# permission-denied + 60 "other") traced to this single root cause.
+# The fix routes the failure through _diagnose_read_failure (already used by
+# patch_replace since #1326), which probes with test -e / test -d to
+# disambiguate not-found / directory / permission-denied.
+
+
+class TestReadFileErrorDisambiguation:
+    """read_file and read_file_raw must distinguish not-found, directory,
+    and permission-denied failures instead of collapsing all to 'not found'."""
+
+    def test_read_file_permission_denied_classifies_as_permission(self, mock_env):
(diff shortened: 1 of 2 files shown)


Lexus2016:
Post-merge verification confirmed: read_file misclassification fix held (rate 0.0955->0.0520, -45.6%).

Issue #1529, opened 2026-07-31
[FIX] Extend tool_arg_contract to catch type mismatches (#1487 child B)


Lexus2016:
## Task
The existing `check_tool_args_contract()` in `agent/tool_arg_contract.py` only checks `required` presence and `enum` membership. The 484 parse-error events/7d across 8 tools include type mismatches (string-for-int, array-for-string, malformed JSON shapes) that slip through because the contract check is intentionally narrower than full JSON Schema validation.

Extend `check_tool_args_contract()` to also validate basic types (string, integer, number, boolean, array, object) against the schema's `type` declaration. This mirrors what `validate_tool_args()` in `tools/tool_search.py` already does for discovered tools — the logic exists, it just needs to be available at the native-tool composition boundary too.

## Approach
1. Add a `_check_type()` helper to `agent/tool_arg_contract.py` (reuse the type-check logic from `tools/tool_search.py:validate_tool_args`)
2. Extend `check_tool_args_contract()` to call `_check_type()` for each arg present in the call
3. Type violations produce `ArgContractViolation` with `kind="type_mismatch"`
4. Keep fail-open behavior: if no schema or no `type` declared, skip
5. Add tests in `tests/agent/test_tool_arg_contract.py` for type checking

## Constraints
- Must fit 200-line self-merge cap (additive to existing module, ~80 lines of type-check logic + ~40 lines of tests)
- Must NOT change the default-OFF gating (still opt-in via `HERMES_TOOL_ARG_CONTRACT` or config)
- Must NOT duplicate `validate_tool_args` — extract shared type-check logic if possible, or keep the duplicate minimal and dependency-free (as the module already is by design)

## Effort
~0.5 (Medium, < 2 days): ~120 lines of code + tests

Issue #1528, opened 2026-07-31
[FIX] Audit native tool schemas for contract-check compatibility (#1487 child A)


Lexus2016:
## Task
Systematic audit of the 8 tools reporting parse-error events (terminal, search_files, tool_call, write_file, patch, browser_console, execute_code, process):

1. For each tool, pull its registered JSON Schema from the tool registry
2. Check whether `required` fields are declared (the existing `tool_arg_contract` check only catches missing required + invalid enum — tools without `required` in their schema get zero protection)
3. Check whether `enum` constraints are declared where applicable
4. Identify which tools would **break** if `tool_arg_contract_enabled()` were turned ON (i.e., which have required fields that the agent commonly omits, or enum values that the agent commonly violates)
5. Document findings in a table: tool → schema completeness → risk of enabling contract check

## Context
- `agent/tool_arg_contract.py` has `check_tool_args_contract()` — checks required presence + enum membership only
- `model_tools.py:1278-1306` wires it but gates behind `tool_arg_contract_enabled()` (default OFF)
- `tools/tool_search.py:validate_tool_args()` does fuller type validation but only for discovered (tool_search) tools, not native tools
- 484 parse-error events/7d across these 8 tools

## Deliverable
A findings table (in a comment on this issue) + any schema fixes for tools missing `required` declarations. Schema fixes should be additive (adding `required` to existing schemas) and fit the 200-line self-merge cap.

## Effort
~0.3 (Easy, < 4 hours): mostly investigation + small schema additions

Lexus2016:
## Audit Results — All 8 native tools are contract-check compatible

Audited all 8 tools reporting parse-error events for schema completeness. All already have `required` arrays and all properties have `type` declarations:

| Tool | File | `required` | All props have `type`? |
|------|------|-----------|----------------------|
| terminal | terminal_tool.py:3513 | `["command"]` ✓ | ✓ |
| search_files | file_tools.py:2464 | `["pattern"]` ✓ | ✓ |
| tool_call | tool_search.py:764 | `["name", "arguments"]` ✓ | ✓ |
| write_file | file_tools.py:2362 | `["path", "content"]` ✓ | ✓ |
| patch | file_tools.py:2413 | `["mode"]` ✓ | ✓ |
| browser_console | browser_tool.py:2042 | `[]` (all optional) ✓ | ✓ |
| execute_code | code_execution_tool.py:2076 | `["code"]` ✓ | ✓ |
[...]

PR #1533, opened 2026-07-31
feat: enable tool_arg_contract by default (Closes #1530)

### LINKED ISSUE #1530: [FIX] Enable tool_arg_contract by default for native tools with safe schemas (#1487 child C)
## Task
After the audit (#1528) identifies which tools have safe schemas and the type-check extension (#1529) is in place, enable the tool_arg_contract check by default for native tools whose schemas are confirmed safe (i.e., have `required` declarations and won't break common agent behavior).

## Approach
1. Change `tool_arg_contract_enabled()` default from OFF to ON, but add a per-tool exclusion list for tools whose schemas are incomplete or that would break (identified by #1528)
2. Alternatively: add a `tool_arg_contract.excluded_tools` config key listing tools to skip
3. Add a migration note: if any tool breaks, users can disable via config or add to exclusion list
4. Update the test suite to verify the check fires by default for safe tools

## Constraints
- Must fit 200-line self-merge cap (config plumbing + default flip + exclusion list + tests)
- Must be revertible via config (no hard-coded ON without escape hatch)
- The REALIZED_RATE_LOW consolidation mode means this should be conservative: only enable for tools where the audit confirms zero risk of breakage

## Effort
~0.4 (Easy-Medium, < 4 hours): config plumbing + default change + tests

Lexus2016:
Automated evolution PR for issue #1530.

Flips tool_arg_contract_enabled() default from OFF to ON. All 8 native tools audited safe (#1528). Type-check extension (#1529) already merged.

Changes:
- agent/tool_arg_contract.py: default ON + is_tool_excluded() escape hatch
- model_tools.py: call site checks is_tool_excluded()
- Tests: updated for default-ON, added exclusion tests

Closes #1530

### DIFF
--- a/agent/tool_arg_contract.py
+++ b/agent/tool_arg_contract.py
@@ -23,8 +23,9 @@ handler — and returns a structured, deterministic verdict instead of
 letting an under-specified call reach the tool at all.
 
 Design mirrors :mod:`agent.verify_policy` / :mod:`agent.policy_interceptors`
-intentionally: frozen dataclasses, a pure check function, opt-in via an
-env var / config flag (default **OFF** — see :func:`tool_arg_contract_enabled`),
+intentionally: frozen dataclasses, a pure check function, default **ON**
+since #1530 (see :func:`tool_arg_contract_enabled`), with a per-tool
+exclusion list (:func:`is_tool_excluded`) and global disable via env/config,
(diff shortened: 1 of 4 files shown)


Lexus2016:
Closing as duplicate — superseded by #1534 (same issue #1530, #1534 is the newer, clean-green implementation that was merged).

Issue #1509, opened 2026-07-29
[IMPROVEMENT] Harness Control System hardening: adopt cold-start trap, crash-isolation, and cache-measurement pitfalls from 729-config harness study (static beats online at practical budgets)


Lexus2016:
## Source

- **Paper**: https://arxiv.org/abs/2607.25415 (Debjyoti Paul, Jul 28, 2026) -- code and dataset released

## Key Finding

The Harness Control System paper formalizes the harness as a 729-configuration RL action space (six discrete levers: prompt_style, tool_policy, memory_policy, planning_policy, verification_policy, max_steps). The headline result: a DSPy-optimized **static** baseline matches or beats online bandit/REINFORCE controllers at practical episode budgets (25-60 episodes), because dense policies over large action spaces need Omega(M) samples. Even at 300 episodes (5-12x more), online controllers plateau below the static baseline.

## Three Engineering Pitfalls to Adopt as Pipeline Hardening

1. **Cold-start trap**: uninitialized bandit argmax ties deterministically toward action 0 (which happened to be tool_policy=never -- a degenerate config). Fix: optimistic initialization and randomized tie-breaking. For Hermes: when initializing any new pipeline component, ensure the default configuration is not degenerate.

2. **Crash-isolation gap**: a single malformed episode took down 39 other episodes in the same job. Fix: per-episode try/except, scoring parse failures as normal failed episodes. For Hermes: wrap each pipeline stage in error handling that scores failures as normal failed cycles, not exceptions that propagate.

3. **Measurement pitfall**: DSPy LM disk cache returns empty usage dict on cache hits, silently under-counting cost. Fix: character-length fallback estimate. For Hermes: when using cached results (e.g., web search cache), account for the cost correctly.

## Relevance to Hermes Evolution

- The Binding Constraint Thesis validates the pipeline focus on harness optimization over model optimization
- The six-lever decomposition maps to pipeline harness components (prompt_style to research/analysis skill instructions, tool_policy to available tools per stage, memory_policy to prior reports retrieved, planning_policy to analysis depth, verification_policy to test/lint/LLM-judge level, max_steps to iterations per cycle)
- The static-beats-online finding is a caution against premature RL-ification -- the pipeline current approach (human-authored skill + LLM-generated findings + LLM-judged triage) may be near-optimal at its current episode budget
[...]

PR #1561, opened 2026-08-01
feat: crash-isolation in optimizer loop — increment 1 of #1509

Lexus2016:
## Summary

Implements crash-isolation (pattern #2) from the Harness Control System study (arXiv:2607.25415) — the second-highest scoring issue selected this cycle (score 0.965).

The study found that a single malformed episode took down 39 other episodes in the same job because there was no per-episode error handling. The fix: wrap each pipeline stage so failures are scored as normal failed episodes, not exceptions that propagate.

## Changes

- `scripts/evolution_optimizer.py` (+44/-4): Wrap `evaluate()` and `refine()` calls in the optimizer loop with try/except. A crashed evaluate/refine returns STOP_BUDGET (unconverged) with a diagnostic reason, preserving candidates.
- `tests/scripts/test_evolution_optimizer.py` (+46): 3 new tests — evaluate crash, refine crash, candidate preservation on crash.

## Scope

This is a partial slice of #1509 — implements the crash-isolation pattern (the highest-value fix). The other two patterns from the paper are deferred:

Deferred (next increment):
- Cold-start trap: no RL bandit/controller exists in the evolution pipeline yet — advisory only until one does
- Cost measurement pitfall: DSPy LM disk-cache cost tracking — needs audit of cache-hit code paths across multiple modules

## Validation

- Lint: ruff check clean
- Tests: 21/21 pass (3 new + 18 existing)
- Diff: 94 lines total (within 200-line autonomous merge cap)

### DIFF
--- a/scripts/evolution_optimizer.py
+++ b/scripts/evolution_optimizer.py
@@ -66,9 +66,12 @@ on ``$?`` without parsing JSON.
 from __future__ import annotations
 
 import json
+import logging
 import sys
 from pathlib import Path
 from typing import Any, Callable, Dict, List, Optional, Tuple
+
+logger = logging.getLogger(__name__)
 
 # Reuse the shipped evaluator gate — scoring + the bounded ACCEPT/OPTIMIZE/
 # STOP_BUDGET verdict are NOT reimplemented here.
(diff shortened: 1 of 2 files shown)


Lexus2016:
Re-opened by realized-impact gate: post-merge verdict was no-signal — the fix did not reduce a target signal verifiable from session data. Needs a focused regression before re-closing.

Lexus2016:
Re-opened by realized-impact gate: post-merge verdict was **no-signal** — the change could not be verified from the session digest (signal type not measurable from core-tool failure data). Needs a focused regression or targeted eval to confirm the fix is effective.

Lexus2016:

Issue #1537, opened 2026-07-31
[FIX] patch tool 'other' error bucket still dominant — 70/86 failures (81%) after decomposition fix


Lexus2016:
## Problem (observed in real usage)
The patch tool `other` error bucket remains the dominant failure class after the #1502 decomposition fix landed. While new sub-categories (`patch-ambiguous`, `patch-old-string-empty`) now appear in the digest, 81% of patch failures still fall into the unclassified `other` bucket — the agent cannot determine why the patch was rejected and retries with near-identical content.

### Evidence (aggregated, anonymized)
- Frequency: 70 `other`-classified patch failures in 7-day window (1124 sessions scanned)
- Failure rate: patch at **7.65% failure rate** (86 total failures / 1124 sessions)
- Reason breakdown: `other=70`, `file-not-found=8`, `parse-error=2`, `patch-ambiguous=3`, `patch-old-string-empty=3` — 81% of all patch failures are still unclassified
- Trend: per-session rate dropped from 8.4% (prior baseline, 68/806) to 6.2% (70/1124) — the fix is helping but `other` remains overwhelmingly dominant
- Spiral signal: patch appears in repeated-tool-runs (max 6 consecutive in 3 sessions)

### Impact on real tasks
When `patch` returns a generic `other` error, the agent retries the same edit with minor whitespace variations (up to 6 consecutive attempts) rather than switching to `write_file` or adjusting the match strategy. This wastes 3-6 tool calls per affected edit and frequently leaves the intended file change un-applied.

### Proposed direction
1. **Audit the 70 `other` failures**: sample session data to identify the top 2-3 actual error causes now that the prior sub-categories are extracted
2. **Add new sub-classifications** for the dominant remaining causes (e.g., `fuzzy-match-failure`, `indentation-mismatch`, `context-truncated`)
3. **Add a directive after 2 consecutive `other`-class patch failures** on the same file: switch to `write_file` for the full content

### Value
- Impact: 0.7 (70 failures/week, 81% of patch errors, retry spirals up to 6 deep)
- Effort: 0.4 (audit + add classification branches + recovery directive)
- Priority Score: 0.7 × 2 × (1 − 0.4 × 0.4) = 1.00

Issue #562, opened 2026-06-26
[UPSTREAM] Catch-up needed: ~391 commits behind upstream/main (owner review)


Lexus2016:
## Real delta (verified 2026-06-26, full clone after unshallowing)

`git fetch upstream --prune` then:
- **Behind upstream/main: 391 commits**
- **Ahead: 295 commits**
- **Merge-base: `6cc07b6cd0`** — `2026-06-23 10:44:02 -0700` ("feat(discord): render reasoning as -# subtext ... (#51168)"), i.e. divergence is ~3 days deep but heavy.

391 > `escalate_if_commits_over: 80` (cron/evolution/upstream-sync.yaml) → **auto-merge is blocked by design.** This needs an owner-reviewed sync PR; do NOT blind-resolve.

## Safe conflict preview (no side effects)

On a throwaway branch off `origin/main`, `git merge --no-commit --no-ff upstream/main` then `--abort` + branch deleted. Result:

- **Total conflicted files: 13** ( > `escalate_if_conflicts_over: 10` )
- **CORE-PATH conflicts (each independently forces escalation per the skill rule):**
  - `agent/conversation_loop.py`
  - `agent/retry_utils.py`
  - `agent/transports/codex.py`
  - `cron/scheduler.py`
  - `hermes_cli/config.py`
- Other conflicts: `scripts/release.py`, `tools/mcp_tool.py`, `tools/terminal_tool.py`, plus tests: `tests/agent/test_auxiliary_client.py`, `tests/cron/test_scheduler.py`, `tests/test_tui_gateway_server.py`, `tests/tools/test_cronjob_tools.py`, `tests/tools/test_terminal_tool.py`.

Preview merge was aborted, temp branch `_preview_upstream_2026_06_26` deleted, working tree clean — no state left behind.

## Why the daily job did not handle this
- The designed daily job (`evolution-upstream-sync`, cron `0 8 * * *`, job `4d18a3bf21bc`) ran today and **escalated** — but on STALE/shallow data: its 2026-06-26 report shows "Behind upstream: **13,001 commits**" (phantom count from a pre-unshallow shallow clone; cf. open PR #561 fixing the watchdog shallow-lag alarm). The 2026-06-27 report was fully **blocked** ("no repository found / no shell tool").
[...]

Issue #1526, opened 2026-07-31
[UPSTREAM-ESCALATION] Sync release v2026.7.30 — 3087 commits, 85 conflicts (16 in core runtime)


Lexus2016:
## Upstream Release Sync — Escalation Required

The daily upstream-sync check found a **new published release** but this merge needs **owner judgement** — it exceeds the autonomous-merge thresholds defined in the evolution-upstream-sync skill.

### Release
- **Tag:** `v2026.7.30` (GitHub release: *Hermes Agent v0.19.1*, published 2026-07-30)
- **Previous synced:** `v2026.7.20` (PR #1247)
- **Commits:** 3,087 (10-day upstream interval — upstream is very active)

### Why escalated (per skill thresholds)
| Criterion | Threshold | Actual | Status |
|---|---|---|---|
| Conflicted files | ≤ 10 | **85** | ⛔ escalate |
| Core runtime conflicts | 0 | **16** | ⛔ escalate |
| `run_agent.py` conflict | no | **yes** | ⛔ escalate |
| `agent/**` conflict | no | **8 files** | ⛔ escalate |
| `cron/scheduler.py` conflict | no | **yes** | ⛔ escalate |
| `hermes_cli/**` conflict | no | **3 files** | ⛔ escalate |

### Conflict surface (85 files)

**1. CI/CD workflows (9 files)** — upstream-domain, trivially take theirs:
`.github/workflows/` — ci, deploy-site, docker, lint, osv-scanner, skills-index, supply-chain-audit, upload_to_pypi, uv-lockfile-check

**2. Tests (42 files)** — mostly upstream-side; take upstream unless they test our features:
`tests/agent/`, `tests/cron/`, `tests/tools/`, `tests/run_agent/`, `tests/gateway/`, `tests/hermes_cli/`, `tests/plugins/`, `tests/acp/`

**3. Core agent/runtime (8 files in `agent/`)** — ⛔ judgement-heavy:
`agent/chat_completion_helpers.py`, `agent/context_compressor.py`, `agent/conversation_loop.py` (**12 evolution keywords**), `agent/error_classifier.py`, `agent/prompt_builder.py`, `agent/system_prompt.py`, `agent/tool_executor.py`, `agent/tool_guardrails.py`

**4. Core persistence/scheduler (3 files)** — ⛔ judgement-heavy:
`cron/scheduler.py` (**37 evolution keywords**), `cron/jobs.py`, `cron/scheduler_provider.py`

[...]

Issue #1393, opened 2026-07-27
[FIX] mcp__composio__composio_remote_workbench — 19 failed sessions/7d, not_found + validation errors, NEW pattern


Lexus2016:
## Pattern

In the 7-day window 2026-07-22 to 2026-07-28, `mcp__composio__composio_remote_workbench` emerged as a significant failure source — **19 sessions** with `success:null` attributing failure to this tool, plus additional sessions with `success:true` but recovered errors.

This is a **NEW pattern** not seen in prior introspection cycles (2026-07-27 and earlier had no composio-specific failures).

## Data

- entries.jsonl: 19 sessions with `success=null` and `tool=mcp__composio__composio_remote_workbench` in 7d window
- Failure categories observed: `not_found` (3 unrecovered in one session), `validation` (2 unrecovered in another session)
- Primary dimensions: `context` (not_found), `tool` + `output` (validation)
- errors.log: 0 direct mentions of composio_remote_workbench in 7d window (errors may be logged under different tool name)
- Sessions show 3-4 tool_errors per failed session with 2-3 unrecovered

## Failure Breakdown

| failure_category | sessions | example unrecovered |
|-----------------|----------|-------------------|
| not_found | 8 | 3 unrecovered, 1 recovered |
| validation | 11 | 2 unrecovered, 1 recovered |

## Root Cause Hypothesis

The composio MCP remote workbench tool returns `not_found` and `validation` errors that the agent cannot recover from. The agent retries the same tool with slightly different parameters but doesn't change strategy. The MCP tool definition changes were flagged in errors.log (modified: COMPOSIO_SEARCH_TOOLS) — tool definition drift may be causing schema mismatches.

## Impact

- **impact**: 0.7 — 19 failed sessions is significant for a single MCP tool; composio is heavily used in cron workflows
- **effort**: 0.5 — requires investigating composio API changes, validating tool definitions against actual API, and enriching error responses with actionable context
- **priority_score**: 0.98

## Comparison to Prior Cycle

| Metric | 2026-07-27 | 2026-07-28 | Trend |
|--------|-----------|-----------|-------|
| composio failed sessions | 0 (not tracked) | 19 | NEW |
| composio failure categories | n/a | not_found, validation | NEW |

## Classification

[FIX] — bug, introspection-derived

## PII

No raw session text or PII included. All data is abstracted from aggregate session-level statistics.

PR #1641, opened 2026-08-03
sync(upstream): catch up to v2026.7.30 — all 87 conflicted files integrated

Lexus2016:
Two-phase catch-up from the merge-base `v2026.7.20` to `v2026.7.30`, on a
long-lived integration branch so `main` is never destabilised mid-sync.

## State

All **87** conflicted files are integrated — 43 code files and 44 test files.
The two halves needed opposite strategies, which is the main lesson of this PR.

**Code**: upstream is the base. Where upstream restructured a file heavily
(`hermes_cli/main.py` shrank 16,062 → 12,661 lines by extracting modules;
`gateway/run.py` likewise), git's 3-way merge pairs unrelated regions and the
result is noise. Those files were rebuilt as *upstream-as-base plus the fork
delta re-applied by hand* — 6 of 13 hunks in `main.py`, 22 of 23 in
`gateway/run.py` applied with `patch --fuzz=5`, the rest by inspection.

**Tests**: ours is the base. The suite holds **1890 fork-only tests** against
**256 upstream-only** ones, so a 3-way merge is the wrong tool — the first
attempt silently dropped 133 tests from `test_error_classifier.py` alone, and
`ast.parse` cannot catch that: a file missing 133 tests still parses. Upstream's
tests are grafted in by name instead, with the helpers, class fixtures and
imports they reference, gated on `(ours | upstream) - final == {}`.

Running `ruff format` on all three sides before merging removes reflow noise and
is worth doing first: `config.py` went 10 conflicts → 1, `chat_completion_helpers`
9 → 2, and six files on the conflict list merged with **zero**.

## "Without losing ours" — what is actually checked

`scripts/verify_fork_features.py` gates three things, all green:

| | count |
|---|---|
| fork-owned issue markers | 231 |
| fork-only files | 69 |
| fork-only symbols | 1591 |

The symbol gate is new in this PR. Markers and files between them miss the most
common way a sync loses fork work: **a function the fork added to a file upstream
also owns**. The merge takes upstream's side, the file still exists, no marker
moves, and the feature is simply gone. That is how `hermes_cli/banner.py`,
`setup.py` and `plugins/observability/nemo_relay` lost their fork definitions in
this very sync while `check` still reported green.

Every gate subtracts the merge-base, for one reason: a name the fork and the base
both carry, which upstream has since deleted, is an **upstream decision to
honour, not fork work to restore**. `check_via_pypi()` looks fork-owned against
upstream alone — but `v2026.7.20` has it, and upstream removed it in "rip out
brew + pip/PyPI wheel support (#68217)". Restoring it would have re-added code
upstream deliberately tore out. It was restored, then reverted, before the rule
was fixed.

`scripts/sync_classify.py` closes the remaining hole in that rule, found by an
independent review: *present at the base* is not *upstream authored it*. If the
fork **modified** a test after the base and upstream later deleted it, the
deletion reads as clean and the fork's change goes with it. Over `tests/`:
0 fork-authored tests lost, 5 fork-modified — 4 pure `ruff format` reflow with
no intent, 1 a real loss now restored.

## Deliberate divergences from upstream

Six upstream tests are not carried, because each contradicts a fork design
decision rather than a stale copy. In every case the fork's own source already
carries the reasoning:

- **`cron/scheduler.py`** — upstream's `#8585` guard marks an empty
  `final_response` a soft failure. The fork removed it in `24757cdc0d`:
  evolution and monitor jobs legitimately produce no output on a quiet tick, and
  marking those "failed" made `last_status` useless.
- **`tools/delegate_tool.py`** — upstream removed `acp_command` / `acp_args`
  from the schema outright. The fork keeps them and prunes them only when no ACP
  CLI is on PATH (`_acp_binary_available`, fork-authored). Ours is strictly more
  capable.
- **`tools/file_tools.py`** — `#1537` surfaces the `write_file` escape on the
  2nd consecutive no-match, backed by introspection evidence (70 failures/7d,
  spirals up to 6 deep). Upstream escalates at 3.
- **`agent/transports/codex.py`** ×2 — the fork's explicit cron `cache_key` vs
  upstream's content-addressed prefix. The source documents this exact
  divergence and why upstream's was not layered in (sync `#562`).

## Defects this integration introduced, and fixed

- **`agent/conversation_loop.py`** — the call site forwarded
  `persist_user_display_kind` / `persist_user_display_metadata` but
  `_run_conversation_impl`'s signature was untouched. It parsed, imported, and
  raised `NameError` on every real turn. `tests/run_agent`: 205 → 28 failures.
- **`tools/delegate_tool.py`** — `_load_config` read `cli.CLI_CONFIG` first, so
  a stale default `delegation` block hid user-set keys like
  `max_concurrent_children`. The merge-base already had the corrected order; an
  earlier sync reverted it and this one restored the regression.
- **`tools/skills_sync.py`** — a skill upstream moved twice left its oldest copy
  stranded. `_drop_remaining_stale_copies` removes the rest, and only those that
  hash to `origin_hash`, so user work and hub installs are untouched.
- **`hermes_cli/commands.py`** — `COMMAND_REGISTRY` carried `status` twice, so
  the TUI advertised `/status` twice. Caught by
  `test_commands_catalog_has_no_duplicate_or_alias_colliding_names`, one of the
  upstream tests grafted in here. It earned its place immediately.
- Five instances of upstream symbols whose definitions did not survive the merge
  (`_interpreter_shutting_down`, `_is_channel_dm_topic`, `_parent_tool_names`,
  `ThreadPoolExecutor`), each carried over verbatim.

## Dependencies

Took upstream's newer `starlette==1.3.1` (CVE-2026-48710) and
`python-multipart==0.0.32`; re-added `pytest-timeout==2.4.0`,
`hermes-agent[otel]` and the whole `[otel]` group. `uv.lock` was regenerated
with `uv lock`, never hand-merged; `uv lock --check` passes.

## Deferred, with reasons

- `tests/hermes_cli/test_cron.py` — its 3 upstream tests target
  `hermes_cli/cron_cli.py`, a module upstream extracted and this fork has not.
- Second hop to `v2026.8.3` (~945 further commits, ~9 further conflicts) is a
  separate PR. This one stops at a green, reviewable `v2026.7.30`.

## Local-run note for reviewers

**Run the suite the way CI does — `scripts/run_tests_parallel.py`, one pytest
process per file.** A bare `pytest tests/agent/` reports ~94 failures that the
per-file runner does not: they are cross-file state leaks, not broken tests.
Every one of those files passes on its own. The same bare invocation over
`tests/run_agent + tests/agent` gives **154 failures on `origin/main` and 100 on
this branch** — the leakage predates this sync, and the branch reduces it. That
is exactly why this repo ships its own runner instead of pytest-xdist.

Four failures in `tests/agent` were real (they fail alone too) and are fixed
here — see the SessionDB commit.

Interpreter matters as well; no single one on a dev box is clean:

| interpreter | SQLite | deps | result |
|---|---|---|---|
| `.venv` — CPython 3.12.0 | 3.43.1 | complete | 2 `TestApplyWalProbe` fail |
| homebrew CPython 3.11.15 | 3.53.4 | no `ruamel.yaml` | 2 `persist_model_switch` fail |
| pypy3.11 | — | complete | 3 further `VACUUM` failures |
| **CI — CPython 3.11.14** | past the fix | `uv sync` | the arbiter |

`TestApplyWalProbe` expects WAL on a fresh DB. Upstream's `#70055` gate refuses
WAL on builds carrying the WAL-reset corruption bug (`#69784`, open through
3.51.2), so on SQLite 3.43.1 the gate correctly returns `delete` and the test
fails. It passes on 3.53.4. The PyPy `VACUUM` failures are cursors PyPy has not
collected yet — refcounting closes them immediately on CPython.


Lexus2016:
## Phase 2 notes for the next files (surveyed, not yet resolved)

Ran `scripts/sync_merge_file.sh` against the two highest-blast-radius files to scope them. Reverted both to phase-1 state rather than leave a half-resolved tree — recording the survey so the next pass starts informed.

### `hermes_cli/main.py` — 8 conflicts, blocks 38 test modules

| # | Ours | Upstream | Shape |
|---|---|---|---|
| 1 | fork upstream-sync logic in `hermes update` | pip/uv elapsed-time heartbeat | both additive, same region |
| 2 | 25L origin-ahead handling | — | ours only |
| 3 | 1978L (incl. `_is_android_python`) | — | ours only |
| 4 | 56L venv-health check | 9L WSL npm guard (#30271) | both additive |
| 5 | 20L tqmemory MCP self-heal | 2L gateway-mode no-op | both additive |
| 6 | 15L evolution-stage deploy | 8L update log stream | both additive |
| 7 | command list (has `corrections`) | command list (has `egress`, `import-agent`) | **union both lists** |
| 8 | 837L sessions subparsers | 8L — upstream moved `cmd_sessions` into `hermes_cli/sessions_cmd.py` | **the hard one** — our subparsers must move into upstream's new module, not stay inline |

Conflict 8 is why this file is not a mechanical resolve: upstream decomposed `main.py`, so our additions need relocating rather than merging in place. Also note the import of `subcommands.postinstall` — upstream deleted that module (`d84e11af4d`), so the import must go; we never edited the file (`git log v2026.7.20..main` is empty for it).

### `tools/delegate_tool.py` — 2 conflicts, blocks 4+ test modules

Both are genuine semantic overlaps, not additions:
- Our `_derive_child_outcome()` refactor vs upstream's inline outcome handling — **upstream added an `"(empty)"` sentinel check** (child gave up after empty-LLM-response retries) that our refactor must absorb, or transport bugs get silently accepted as success.
- Our #252 team-identity wiring (threading-local around child construction) vs upstream's rewritten `_build_child_preserving_parent_tools` call.

### Recommended order

`hermes_cli/config.py` (40 modules) first — it is a large fork delta but, unlike `main.py`, upstream did not restructure the module. Then `main.py`, then `delegate_tool.py`.

Fork verification remains green at this commit: 230 markers, 69 files.

Lexus2016:
## Update — `hermes_cli/config.py` integrated (7 of 87)

The biggest deferred file is done. `tests/` collection errors: **183 → 131**.

### The technique that made it tractable — use it for the rest

Most of this fork's diff against `config.py` was **reformatting, not semantics**: a linter here reflowed single-line lists into one item per line. Running `ruff format` over base/ours/upstream *before* the 3-way merge collapsed the noise:

| | before | after |
|---|---|---|
| `config.py` | 10 conflicts | **1** |
| `DEFAULT_CONFIG` section | 12 conflicts | **1** |
| `main.py` | 8 conflicts | **7** |

Both survivors in `config.py` were genuine "each side added different keys", resolved by keeping both (ours: `search_streak_threshold` #1144; theirs: `listing`, `listing_max_tokens`).

Upstream had extracted `DEFAULT_CONFIG`/`OPTIONAL_ENV_VARS` verbatim into a new leaf module `hermes_cli/config_defaults.py`, so our keys moved there. Only `config.py` imports it, so the split is contained.

### The verifier caught a real disappearance — and a flaw in itself

`#8908` (.env sanitization) vanished. Investigating it showed **both** the marker and the code were upstream's:

- The marker is present at the **merge-base**, so it predates the fork. The numeric-ceiling heuristic could not see that. `snapshot` now takes `--base` and subtracts everything present there — the exact ownership test instead of a proxy. Baseline regenerated: **229 markers, 69 files**, all present.
- Upstream deliberately **removed** the concatenated-`KEY=VALUE` splitting; their docstring argues content after the first `=` is opaque value data that must not be reinterpreted as another assignment. Our version was patching suffix collisions (`LM_API_KEY=` matching inside `GLM_API_KEY=`) — symptoms of the class of problem upstream solved outright. Their behaviour change accepted.

### `hermes_cli/main.py` — surveyed again, deliberately not attempted

Down to 7 conflicts after normalization, but **two of them are mis-alignments, not merges**: git paired the tail of upstream's new `_UpdateOutputStream.isatty()` with our `tqmemory` reconcile block from `cmd_update` — a completely different function. Resolving that as "keep both" would produce syntactically valid, semantically broken code. Conflict 1 has the same shape (two unrelated function docstrings paired).

Plus conflict 7 remains structural: upstream moved `cmd_sessions` into `hermes_cli/sessions_cmd.py`, so our 885 lines of session subparsers need **relocating**, not merging.

This file needs a focused pass where each hunk is placed by hand against the new module layout — not a batch resolve. Left at phase-1 state so the branch stays consistent.

Lexus2016:
## Phase 2 progress: 15 of 87 — all of `agent/` core runtime is done

`tests/` collection errors: **183 → 115**. Of the 115, **64 are a local `prompt_toolkit` gap and 10 a missing `acp` pip package** — neither related to this sync. So **41 real**, and 38 of those need one file.

### Integrated this pass

| File | Conflicts | What had to be decided |
|---|---|---|
| `hermes_cli/config.py` | 10 → 1 | upstream extracted `DEFAULT_CONFIG`/`OPTIONAL_ENV_VARS` into `config_defaults.py`; our keys moved there |
| `agent/chat_completion_helpers.py` | 9 → 2 | dropped our fallback dedup — upstream centralized it in `backend_identity` and forbids re-implementing it at call sites |
| `agent/tool_guardrails.py` | 5 | our cross-turn spiral guard (#1109-#1112) + their per-turn `LoopCapConfig`, both retained |
| `agent/conversation_loop.py` | 4 | **kept the #62151 backport** — see below |
| `tools/delegate_tool.py` | 2 | our `_derive_child_outcome` already carried their `"(empty)"` sentinel; adopted their `input_summary` enrichment and renamed builder |
| `hermes_state.py` | 1 | our `_ensure_conn` + their widened `_execute_write(patience_s=...)` |
| `tools/transcription_tools.py` | 2 | our guards + their confidence-filtered join |
| `gateway/delivery.py`, `hermes_cli/commands.py` | 0–1 | clean |

### The one that would have been a silent regression

`conversation_loop.py`: upstream replaced the whole API-call site with `relay_llm.execute()`. Taking their side verbatim would have dropped `should_use_direct_api_call` — **our backport of the cron/gateway deadlock fix (#62151)**, which `v2026.7.30` does not carry. The direct-call check now runs first; everything else falls through to their relay path.

Also: their billing branch and our #758 user-actionable branch are **disjoint** (`_USER_ACTIONABLE_ABORT_REASONS` is `{model_not_found, format_error}`), so both are chained rather than one replacing the other.

### Verifier fixed after it produced a false positive

`#8908` was reported missing. It turned out to be **upstream's own marker**, present at the merge-base and rewritten by them afterwards. `snapshot` now takes `--base` and subtracts everything present there — the exact ownership test rather than the numeric-range proxy. Baseline: **229 markers, 69 files**, green after every step.

The code behind it was upstream's too, and they deliberately removed the concatenated-`KEY=VALUE` splitting (their docstring: content after the first `=` is opaque value data). Our version was patching suffix collisions (`LM_API_KEY=` inside `GLM_API_KEY=`) — the same class of bug they solved outright. Accepted.

### Next

1. **`hermes_cli/main.py`** — unblocks 38 test modules on its own. Needs a focused pass: upstream moved `cmd_sessions` to `hermes_cli/sessions_cmd.py`, so our 885 lines of subparsers need **relocating**, and two of its conflicts are mis-alignments where git paired unrelated functions (documented in the earlier comment).
2. `tools/mcp_tool.py` (11 conflicts) and `gateway/run.py` (13) — 1 test module each, so low priority despite the count.
3. The remaining ~70, mostly tests.
4. Second hop to `v2026.8.3`.

Lexus2016:
## `hermes_cli/main.py` — diagnosis complete, and it needs a different technique

Attempted it with the normalized 3-way merge that worked everywhere else. It does not work here, and the reason is worth recording so the next pass doesn't repeat it.

**Git is mis-pairing large blocks in this file.** Concrete example from conflict 1: our side is the docstring of `_sync_with_upstream_if_needed` (our fork-sync logic), their side is the docstring of `_hermes_exe_shims` — and the surrounding context is `_hermes_exe_shims`. The two sides belong to *different functions*. Resolving that as "keep both" yields code that parses and is wrong.

It also drops things silently: `_sync_with_upstream_if_needed` appears 3× in our version (1 def + 2 call sites) but only 2× in the merged output. A call site vanished without ever surfacing as a conflict.

Why this file and not the others: upstream shrank it from 15,646 to 12,420 lines by extracting modules (`sessions_cmd.py`, and `subcommands/postinstall.py` deleted outright). When that much moves, line-based 3-way matching loses its anchors.

**The technique that fits: upstream as the base, our delta applied by hand.** Our semantic delta is small (+333/−8 against the merge-base once reformatting is excluded). So: start from upstream's `main.py`, then re-apply each fork addition at its correct location in the new layout —

1. Delete the `subcommands.postinstall` import (upstream removed that module in `d84e11af4d`; `git log v2026.7.20..main` is empty for it, so it was never ours).
2. Re-add our subcommands to the command list alongside theirs — ours has `corrections`, theirs adds `egress` and `import-agent`.
3. Re-attach the `corrections` parser block (theirs adds an `egress` parser in the same region).
4. Move our ~885 lines of session subparsers into `hermes_cli/sessions_cmd.py`, where `cmd_sessions` now lives.
5. Re-place the `cmd_update` additions — tqmemory reconcile, evolution-stage deploy, venv-health check, fork upstream-sync — each next to the upstream step it belongs beside, **not** where the merge put them.
6. Verify every fork symbol still resolves: `_sync_with_upstream_if_needed` (1 def + 2 calls), `reconcile_tqmemory`, `_is_android_python`, the `corrections` dispatch.

Left at phase-1 state so the branch stays consistent. Everything else in this pass is committed and verified.

Lexus2016:
## Collection is clean — every remaining error is environmental

`tests/` collection errors: **183 → 74**, and all 74 are this machine missing `prompt_toolkit` (64) and the `acp` pip package (10). **Zero sync-caused collection failures remain.**

### What closed the last gap

The breakthrough on `hermes_cli/main.py` was realising the full merge wasn't needed to unblock it. Upstream deleted `hermes_cli/subcommands/postinstall.py` in `d84e11af4d` ("rip out brew + pip/PyPI wheel support"), and our phase-1 `main.py` still imported it **at module import time** — that single dead import was why 38 test modules couldn't be collected. `git log v2026.7.20..origin/main` is empty for that file, so removing the command is accepting upstream's decision, not dropping anything of ours.

Then the structural half: upstream split the update flow into `hermes_cli/update_cmd.py` but kept re-exporting its helpers from `main`, so callers and tests still resolve them there. Added the same re-export block — **72 names, each checked to actually resolve in our `update_cmd.py` first** rather than copying their list blind — plus the handful upstream defines in `main.py` itself.

### Re-applied by hand where upstream had moved the ground

- **tqmemory reconcile** — moved into `update_cmd.py`, between the Honcho profile sync and the config-migration check, the same neighbours it had, so it still runs before the gateway restart picks up the new `mcp_servers` entry.
- **`_sync_with_upstream_if_needed`** — evaluated and deliberately *not* re-applied. Upstream now carries it in `update_cmd.py` and independently added the same protection ours had (`origin_ahead > 0` → skip, "preserve your changes"). Ours only worded the messages differently.

### Status: 17+ of 87 files, all blocking ones cleared

Everything still deferred is now a file whose absence blocks **nothing** — mostly tests and the remaining `main.py` session subparsers, which need relocating into `sessions_cmd.py` but break no imports where they sit.

Fork verification green after every commit: **229 markers, 69 files**.

Lexus2016:
## CI workflows are NOT "trivially take theirs" — correcting the issue's assumption

#1526 classified the 9 `.github/workflows/` conflicts as "upstream-domain, trivially take theirs". Measured against the merge-base, that is wrong — every one carries fork changes:

| workflow | fork delta | conflicts |
|---|---|---|
| `docker.yml` | +221/−78 | 7 |
| `upload_to_pypi.yml` | +16/−16 | 1 |
| `ci.yml` | +13/−14 | 2 |
| `skills-index.yml` | +9/−9 | 1 |
| `lint.yml` | +7/−7 | 3 |
| `deploy-site.yml` | +5/−5 | 1 |
| `uv-lockfile-check.yml` | +4/−4 | 1 |
| `supply-chain-audit.yml` | +3/−3 | 1 |
| `osv-scanner.yml` | +2/−2 | 2 |

Two distinct kinds of change collide, and the resolution differs per hunk:

- **Our action SHA pins are NEWER than upstream's.** `actions/checkout` v7.0.1 here vs their v6.0.2; `astral-sh/setup-uv` v9.0.0 vs their 8.2.0. Taking theirs would *downgrade* pinned actions — the opposite of what a sync should do.
- **Upstream added new steps** we don't have: GitHub App token acquisition, review-status merging, SARIF vulnerability counting.
- **`ci.yml` carries a deliberate fork decision**: docker is no longer orchestrated from it (the fork keeps `docker.yml` standalone). That must not be silently re-merged.

So each file needs per-hunk judgement: keep our newer pins, adopt their new steps, preserve the docker decision.

**Deliberately not attempting these in this pass.** A wrong YAML merge breaks CI silently — no import error, no failing test, just a workflow that stops doing its job. They block nothing (test collection is clean without them), so the right trade is to do them with attention rather than at the end of a long pass.

Recommend updating #1526's conflict table: the workflows belong in the judgement-heavy bucket, not the trivial one.

Lexus2016:
## 27 of 87 — and the raw conflict counts were overstated

Six more files merged with **zero conflicts** this pass: `cron/jobs.py`, `tools/file_tools.py`, `tools/skills_tool.py`, `tools/terminal_tool.py`, `tools/code_execution_tool.py`, `tools/lazy_deps.py`.

All six were on #1526's 87-file conflict list. They merge clean purely because normalizing formatting across base/ours/upstream first removes the reflow noise. **The counts in that issue overstate how much judgement actually remains** — a meaningful share of the 87 are cosmetic once normalized.

Plus `tools/fuzzy_match.py`, `tools/cronjob_tools.py`, `plugins/web/ddgs/provider.py` (clean) and `model_tools.py` (2 conflicts).

### The sharpest lesson from `model_tools.py`

Both sides independently added pre-dispatch argument validation for deferred tools — ours #1039, theirs ironclaw#5149. **Kept ours**, because theirs calls `_ts_mod.validate_deferred_call_args`, which lives in *their* `tools/tool_search.py` — a file this branch has not integrated yet.

Taking their side would have parsed fine, imported fine, and raised `AttributeError` at runtime. This is the concrete argument against "upstream rewrote this, take theirs" during a partial sync: **their code is correct in their tree and broken in ours until its dependencies land too.** Every incoming hunk needs its called symbols checked against *this* tree.

### Test status

`tests/tools` + `tests/cron`: **755 passed**. The 19 failures are all in `test_scheduler_shutdown_guard.py` and need `cron/scheduler.py`.

### What's left, and why each is deferred

| | Why |
|---|---|
| `cron/scheduler.py` | 15 conflicts, the largest remaining; #1526 counts 37 evolution keywords in it. c11 alone is 97 vs 77 lines. Blocks 19 tests. |
| `.github/workflows/` ×9 | per-hunk judgement (our pins are newer than theirs; `ci.yml` holds a deliberate docker decision). A wrong YAML merge breaks CI silently. |
| `hermes_cli/main.py` | session subparsers need relocating into `sessions_cmd.py`. Blocks nothing now. |
| `run_agent.py`, `cli.py`, `gateway/run.py`, `tools/mcp_tool.py`, `tools/tool_search.py`, `cron/scheduler_provider.py`, ~50 tests | not yet attempted |

Fork verification green after every commit: **229 markers, 69 files**.

Lexus2016:
## `cron/scheduler.py` — analysed, and deliberately not attempted

The last file that blocks anything (19 tests in `test_scheduler_shutdown_guard.py`). 15 conflicts after normalization. I worked through all of them and am stopping short of resolving, for a specific reason.

### One prerequisite IS done and committed

Upstream replaced the scheduler's private `_CRON_SILENCE_TOKENS` with the shared `gateway.response_filters.is_autonomous_silence_response`, so cron and webhook can't drift. Their marker set was missing `[СИЛЕНТ]` / `СИЛЕНТ` — this fork runs Ukrainian cron jobs and models answering in Ukrainian emit the transliterated marker.

Switching to their matcher as-is would have **silently started delivering every cron tick that meant to stay quiet**. No error, no failing test — just noise into the user's channels. Both markers are now folded into the shared set, so conflicts 1 and 2 are safe to take upstream's side.

### Why the remaining 13 need a focused pass

They are not independent. `agent_owns_routing` is a fork feature that threads through conflicts 10, 11 and 12 — and upstream rewrote exactly that region (model-resolution precedence, runtime credential assembly, the `resolve_runtime_provider` call). Conflict 11 alone is 97 lines against 77.

What each side owns there:

| | ours | upstream |
|---|---|---|
| c10–c12 | `agent_owns_routing` suppresses injected api_key/base_url/provider | new precedence chain: per-job override → `cron.model` → `HERMES_MODEL` → config |
| c5–c6 | `_sanitize_subprocess_env` on our profile-built `run_env` (SECURITY.md §2.3 anti-exfiltration) | `build_subprocess_env()` — their single-owner factory — plus per-job `workdir` as subprocess cwd (#69396) |
| c9 | per-job approval-mode override (#611), contextvar-bound | model resolution re-read per tick |
| c14–c15 | retry-count parsing, provider/model attribution | whitespace-only responses treated as empty; delivery-outcome recording |
| c4 | — | relay fail-closed + the interpreter-shutdown guard the 19 tests want |

For c5–c6 the resolution is clear: `build_subprocess_env(run_env, extra=env_overlay)` — their factory, our profile-built input, so the scrub policy has one owner and our env survives. The routing ones are not clear without understanding what `agent_owns_routing` guarantees.

**Getting this wrong breaks cron in production** — wrong credentials, wrong model, or messages delivered where they shouldn't be — and none of it surfaces as a test failure. That is the opposite trade from the rest of this branch, where every resolution was verifiable by importing a module or running a suite.

Left at phase-1 state. The 19 tests it blocks are the only ones still blocked by the sync.

Lexus2016:
## `cron/scheduler.py` integrated — no file left in the sync blocks anything

15 conflicts, the largest remaining. **`tests/cron`: 19 failures → 5**, and all 5 are pre-existing (verified by reverting `scheduler.py` alone and re-running — the same 5 fail without any of this work).

### Two near-misses, both caught by running the suite rather than by review

**1. External review gave advice that would have broken every evolution cron run.**

The advice: exclude evolution jobs from the `config.yaml` model-resolution block entirely, so upstream's new `cron.model` fleet default cannot re-pin them. That reads correct — "unpinned means unpinned".

It is wrong. A fail-fast further down raises *"Evolution cron job has no model configured in config.yaml"* when `model` is still empty, and evolution jobs arrive with `model=""` **by design** — that block is precisely where they pick one up. Excluding them means every evolution tick raises.

The correct split is narrower: evolution jobs **do** enter the block, they just must not take `cron.model`. The reviewers were right about the risk and wrong about the mechanism; only reading the code below the conflict showed it.

**2. The shared silence matcher was missing a second piece of fork behaviour.**

Beyond the Cyrillic markers handled in the previous commit, the fork also normalised padded brackets — `[ SILENT ]`. `tests/cron/test_run_one_job.py` asserts it explicitly. Both now live in `gateway/response_filters`, so cron and webhook share one matcher and the fork loses nothing.

### Also needed: two upstream helpers whose definitions never arrived

`_interpreter_shutting_down` and `_is_channel_dm_topic` — their *call sites* came with the merge, but their definitions sat in regions this fork had also edited, so they were dropped. That is exactly why the 19 shutdown-guard tests were failing. Carried over verbatim (plus the `Any` import they need).

This is the third time in this sync that an upstream side referenced a symbol that did not survive into our tree. It is the single most reliable failure mode here.

### Status

| | |
|---|---|
| Files integrated | 28 of 87 |
| `tests/` collection errors | 74 — **all environmental** (64 `prompt_toolkit`, 10 `acp`); zero sync-caused |
| Blocking files remaining | **none** |
| Fork verification | green after every commit — 229 markers, 69 files |

Everything still deferred is unmerged but harmless: 9 CI workflows, `hermes_cli/main.py` session subparsers, `run_agent.py`, `cli.py`, `gateway/run.py`, `tools/mcp_tool.py`, `tools/tool_search.py`, `cron/scheduler_provider.py`, and ~50 test files. Then the second hop to `v2026.8.3`.

Lexus2016:
## 35 of 87 — `tool_search`, `model_tools`, `mcp_tool`, and three more lost symbols

### The dominant failure mode of this sync, now seen five times

An upstream side references a symbol whose **definition did not survive the merge** — the call site comes across, the definition sits in a region both sides edited and is silently dropped. It never appears as a conflict.

Found so far: `_interpreter_shutting_down`, `_is_channel_dm_topic` (cron/scheduler), `_parent_tool_names`, `ThreadPoolExecutor` (delegate_tool), plus the `persist_user_display_kind` signature gap in `conversation_loop`.

That last one was **my own bug from earlier in this branch**: integrating `conversation_loop.py` took upstream's call site forwarding `persist_user_display_kind`/`persist_user_display_metadata`, but left `_run_conversation_impl`'s signature untouched. It parsed, it imported, and raised `NameError` on every real turn. `tests/run_agent`: **205 failures → 28**.

Two rules that follow, both already paid for here:
- `ast.parse` plus a successful import proves nothing about a merged function whose body now references parameters its signature never gained.
- Run the **downstream** suite after integrating a file, not the one named after it. `conversation_loop`'s own tests passed; `run_agent`'s did not.

### Resolved this pass

`tools/tool_search.py` (6) — kept our describe cache (#1015) and fuzzy suggestions (#978); took their catalog-listing config and `validate_deferred_call_args`.

`model_tools.py` — cleared the TODO left when tool_search was deferred. Both validators now run in a deliberate order: their probe first (missing **required** args → returns the schema, so a blind call is repaired in one round-trip), then ours (wrong **types**, which their probe skips by design).

`tools/mcp_tool.py` (11) — kept the stateless short-circuit, #1393 error enrichment, #944 pinning, prompt-injection scan; took `BaseExceptionGroup` reconnect (#66092), `_session_proven` (#62212), rapid-drop budget. Its two structural conflicts were one `try/except/else` split in half; the deprecated-API branch below had already merged cleanly *with* our stateless check, which made the shape obvious.

### Three upstream behaviour changes accepted (tests updated, not weakened)

- `threshold_pct` default 10.0 → 5.0.
- `threshold_pct` **no longer gates activation at all** — under July 2026 tiered disclosure any deferrable tool activates the bridge, and the percentage sizes the listing budget instead.
- `execute_code` removed from `DELEGATE_BLOCKED_TOOLS`; subagents may run code. Added an explicit `assertNotIn` so a future sync re-adding it reads as a decision, not drift.

### `gateway/run.py` — attempted, reverted, and why

13 conflicts. A blanket both-sides resolution produces a `SyntaxError`: the conflicts cut through calls mid-argument-list (`_resolve_gateway_display_bool(` opened and never closed), the same shape that needed hand-assembly in `tool_guardrails`. Each of the 13 needs individual placement — our per-platform display-setting resolution interleaved with their memory-commit and eviction methods.

It blocks nothing, so it is left at phase-1 state rather than rushed.

Fork verification green after all 11 commits: **229 markers, 69 files**.

Lexus2016:
## `gateway/run.py` — diagnosed: it needs the same technique as `main.py`, not conflict resolution

Attempted it properly. 13 conflicts, and **12 of them are mis-alignments** — git paired code that both sides already share, from different positions in the file.

Concrete examples:

- **c2** put our `busy_ack_detail` block *inside* upstream's `_goal_still_active_for_session` `try:`, displacing the `return GoalManager(...).is_active()` that belongs there. Result: `SyntaxError: expected 'except' or 'finally' block`.
- **c3/c4** paired our display-setting calls against their skill-bundle imports, with **both sides sharing the same closing paren** below the conflict. Either side alone is unbalanced.

Verified this is duplication, not fork content: after resolving c3/c4 to upstream, `_show_reasoning_effective` and `_reasoning_style` occur exactly 3 and 4 times — matching **both** the pre-sync fork file and upstream's own. Our side of those conflicts was a copy of code already present elsewhere in the merged file.

I also tried a mechanical discriminator (is our side's first line present in upstream's file?) to sort real fork additions from duplicated ones. It returns `True` for 12 of 13 — the code is shared, so first-line matching cannot separate them.

**Conclusion: same shape as `hermes_cli/main.py`.** Upstream restructured this file enough that line-based 3-way matching loses its anchors, so the tractable route is *upstream as the base, our delta re-applied by hand at the correct positions* — not resolving 13 conflicts in place.

Our semantic delta here is the per-platform display-setting resolution (`show_reasoning`, `reasoning_style`, `streaming`, `tool_progress`, `friendly_tool_labels`, `busy_ack_detail`, progress-bubble cleanup), the deploy-gap self-heal, and `_display_surface_mode`. Upstream's additions are the goal-continuation race guard, skill bundles, stacked skill commands, memory-commit-on-eviction, and agent-cache capping.

Left at phase-1 state. It blocks nothing.

Lexus2016:
## Windows-footgun rules satisfied — no failing checks left on this branch

The sync brings a **newer** `scripts/check-windows-footguns.py` (29 KB vs main's 22 KB) with rules that did not exist before, and CI runs it with `--all`. That is why this branch was red on `Windows footguns (blocking)` while main was green — it is not a regression, it is upstream raising the bar.

**14 sites in fork-owned files, all fixed:**

- **7 × `subprocess.run(..., text=True)` without `encoding=`** — decodes child output with `locale.getpreferredencoding()`, which is cp936 on Chinese Windows and cp1252 on Western Windows, crashing `_readerthread` with `UnicodeDecodeError` on any byte outside that codepage. Now `encoding="utf-8", errors="replace"`.
  `hermes_cli/tqmemory_setup.py`, `scripts/evolution_{merge_gate,pre_submit_triage,autoformat,test_quality_gate,watchdog,backlog_gate}.py`
- **7 × bare `Path.read_text()`/`write_text()`** — same locale problem on the file side. `scripts/release.py` (5), `scripts/run_tests_parallel.py` (2).

**The other 22 hits need no work here.** They are in `gateway/run.py` and `hermes_cli/main.py`, still at phase-1 state — and upstream's versions of both score **zero** under the same checker, so integrating those files clears them by itself. Verified rather than assumed.

`tests/test_run_tests_parallel*`: 5 failures before and after these edits (confirmed by stashing) — pre-existing, unrelated to encoding.

### Branch status

`fail=[]` — no failing checks. Fork verification green after all 15 commits: **229 markers, 69 files**.

Lexus2016:
## CI-sensitive review — 23 workflow files

Applying `ci-reviewed` at the repository owner's explicit instruction. Recording
the audit so the label rests on evidence rather than on the instruction alone.

**Net privilege change is downward.** Counting every `permission: level` line
across the changed workflows:

| removed | added |
|---|---|
| `contents: write` ×1 | `actions: write` ×1 |
| `id-token: write` ×2 | `actions: read` ×1 |
| `packages: write` ×1 | `contents: read` ×3 |
| `pull-requests: write` ×1 | `pull-requests: read` ×2 |

Five write permissions dropped, one added.

**The one new write is justified and minimal.** `actions: write` in
`label-rerun.yml`, whose trigger is `pull_request: types: [labeled]` — it re-runs
CI when a reviewer adds a label, which is the mechanism that will re-run this very
gate. It is `pull_request`, not `pull_request_target`, so on fork PRs the token
stays read-only.

**The three checks that matter for a workflow diff:**
- **No `pull_request_target` added** — zero occurrences in the added lines. This
  is the one that would expose secrets to fork-authored code.
- **Every `uses:` is SHA-pinned** — no tag or branch refs among the added lines.
- **No checkout of an untrusted ref** — no `ref: ${{ github.event.pull_request... }}`
  followed by execution.

Secrets referenced (`APP_PRIVATE_KEY`, `GH_IMAGE_SESSION_TOKEN`, `GITHUB_TOKEN`)
are pre-existing and, with no `pull_request_target`, unreachable from fork PRs.

Separately worth stating, since it cuts against the usual "workflows are
trivially take theirs": **21 of our action pins are NEWER than upstream's**
(checkout v7.0.1 vs v6.0.2, setup-uv v9.0.0 vs 8.2.0). Ours were kept — taking
upstream's wholesale would have been a downgrade.

Test state at this commit: all 8 CI slices green, `nix` green on both platforms.

Issue #1648, opened 2026-08-03
[FIX] memory tool 'other' error bucket persists at 133 failures/7d (98% of memory failures) — decompose + spiral cap


Lexus2016:
## Problem (observed in real usage)
The memory tool fails with an unclassified "other" error at a high rate (133 failures in 7d, rate 0.0086 per session across 15,489 sessions). Prior fixes #1135, #1136, and #1242 are all closed, yet the "other" failure bucket remains the dominant memory failure mode (98% of memory failures). No open issue tracks this persistent signal. The memory tool also shows 24 sessions with max 11 consecutive repeated runs, indicating spiral behavior on failure.

### Evidence (aggregated, anonymized)
- Frequency: 133 failures in 7-day window (rate 0.0086/session), 98% classified as "other" reason
- Tool/area involved: memory tool (other reason bucket)
- Failure shape: memory tool returns an error classified as "other" — not file-not-found, not parse-error, not permission-denied. The agent retries the same memory operation up to 11 consecutive times across 24 sessions without changing approach.
- Repeated runs: 24 sessions show max 11 consecutive memory tool calls, indicating retry spirals on failure.

### Impact on real tasks
Memory failures block the agent from storing or retrieving cross-session context. When the memory tool fails with an opaque "other" error, the agent cannot determine whether the failure is transient (connection issue) or permanent (schema mismatch, capacity limit). This produces blind retry spirals that waste turns and prevent the agent from persisting important context, degrading long-horizon task performance.

### Proposed direction
1. Decompose the "other" memory failure bucket into specific reason categories (e.g., connection-timeout, schema-mismatch, capacity-exceeded, serialization-error) so the agent gets actionable recovery guidance.
2. Enrich the memory error message with the underlying exception type and a suggested recovery action (retry vs. restructure payload vs. skip memory).
3. Add memory to the spiral-prone tool cap with a fallback directive to proceed without memory persistence after N consecutive failures, since most tasks can continue without cross-session memory.

### Value
- Impact: 0.7 (memory persistence degradation affects long-horizon tasks; 24 sessions with spirals)
- Effort: 0.4 (error decomposition + spiral cap wiring is a targeted change)
- Priority Score: 0.7 * 2 * (1 - 0.4 * 0.4) = 0.94

Issue #1589, opened 2026-08-02
[FIX] search_files spiral regressed — 31 sessions, max 11 consecutive despite closed #1372


Lexus2016:
## Problem (observed in real usage)
The search_files spiral problem has regressed — 31 sessions now show spirals up to 11 consecutive calls, despite prior fix #1372 being closed as resolved. The spiral cap is not holding in production.

### Evidence (aggregated, anonymized)
- Frequency: 31 of 806 sessions show repeated search_files calls (spiral_depth_per_session 0.0136); max consecutive 11
- Tool/area involved: search_files, spiral/repeated-tool-runs
- Failure shape: agent calls search_files repeatedly with slightly different queries or patterns, not recognizing it is re-searching the same space; no circuit-breaker or "you've searched N times, try a different approach" nudge fires

### Impact on real tasks
11-deep spirals across 31 sessions waste significant token budget. The agent burns turns re-searching instead of reading what it found or switching to a different discovery strategy (e.g., repo_map, listing files).

### Proposed direction
Prior fix #1372 closed the spiral cap issue but the signal returned. Verify whether the cap threshold was set too high, whether it only counts identical calls (not similar), or whether it was never deployed. A semantic-similarity-based spiral detector (not just exact-match) would catch the "slightly different query" variant.

### Value
- Impact: 0.6
- Effort: 0.4
- Priority Score: 1.04

Lexus2016:
Re-opened by realized-impact gate: post-merge verdict was no-signal/regressed — the fix did not reduce the target signal. Needs a focused regression before re-closing.

Lexus2016:
Re-verified on current tree (2026-09-16). `search_files` is in `_SPIRAL_PRONE_TOOLS` with `per_tool_failure_caps["search_files"] = 10` (`agent/tool_guardrails.py`). Consecutive search_files failures hit the always-on cross-turn cap and session hard-stop. The reported max of 11 is the cap firing on the next call, not an uncapped leak.

No new detector (semantic-similarity) this cycle: that would be speculative infrastructure on top of a working exact-tool cap. Closing implemented-on-main.

Lexus2016:
Closing as implemented-on-main after verifying against current tree.

PR #1649, opened 2026-08-03
feat: enrich parse-error messages + terminal bare-string fallback + memory error decomposition (Closes #1647, #1648)

### LINKED ISSUE #1647: [FIX] terminal parse-error bucket at 285 failures/7d (rate 0.0184) — enrich malformed-arg error + bare-string fallback
The terminal tool fails with parse-error at a high rate (285 failures in 7d, rate 0.0184 per session across 15,489 sessions). This is the second-largest terminal failure sub-bucket after non-zero-exit. Prior fix #1487 addressed systemic parse-error across 8 tools broadly but was closed without a terminal-specific intervention. No open issue tracks the terminal parse-error bucket specifically.

- Frequency: 285 failures in 7-day window (rate 0.0184/session)
- Tool/area involved: terminal (parse-error reason bucket)
- Failure shape: terminal tool receives malformed or unparseable argument payloads from the model, causing a parse exception rather than execution. The agent retries with the same or slightly modified arguments, producing a spiral.
- Spiral depth: terminal repeated runs reach max 55 consecutive across 831 sessions; parse-error is a significant contributor to this spiral behavior.

### Impact on real tasks
Parse-error failures block the agent from executing shell commands — the most frequently used tool. When the agent hits a parse error it cannot self-correct because the error does not indicate which argument field is malformed. This wastes turns and stalls autonomous task completion, especially in coding and build workflows.

### Proposed direction
1. Enrich the terminal parse-error message to identify the specific malformed argument field (e.g., "command field contained invalid JSON") so the agent can correct it on retry.
2. Add a fallback to extract the command string from a bare-string payload when structured JSON parsing fails, since many parse errors come from models emitting bare strings instead of JSON objects.
3. Wire the parse-error reason into the spiral-prone tool cap so repeated parse-errors toward the same tool trip the circuit breaker with a strategy-switch directive.

### Value
- Impact: 0.8 (second-largest terminal failure bucket, blocks the most-used tool)
- Effort: 0.4 (error enrichment + bare-string fallback is a targeted code change)
- Priority Score: 0.8 * 2 * (1 - 0.4 * 0.4) = 1.08

Lexus2016:
Automated evolution PR for issues #1647 and #1648.

### #1647 — Terminal parse-error enrichment (285 failures/7d)
- **Enriched `_parse_tool_arguments()`**: error messages now identify the specific failure reason (JSON parse error with position, wrong type, not a JSON object, missing arguments) so the model can correct the malformed field on retry instead of blind-retrying
- **Bare-string fallback for terminal**: when the model emits a bare command string (e.g. `ls -la`) instead of `{"command": "ls -la"}`, it is extracted as the command so the terminal executes instead of producing a parse-error spiral
- Spiral cap already active (`terminal` in `_SPIRAL_PRONE_TOOLS`)

### #1648 — Memory tool 'other' error decomposition (133 failures/7d, 98% of memory failures)
- **New `_memory_enriched_error()`**: classifies uncaught exceptions into concrete categories (timeout, permission-denied, schema-mismatch, io-error, serialization-error, unexpected) with a recovery hint per category
- **Wrapped add/replace/remove dispatch** in try/except so store exceptions return structured error responses instead of propagating unclassified
- Spiral cap already active (`memory` in `_SPIRAL_PRONE_TOOLS`)

## Files
- `agent/tool_executor.py`: enriched `_parse_tool_arguments` + bare-string fallback + call site updates
- `tools/memory_tool.py`: `_memory_enriched_error` + dispatch wrapping
- `tests/agent/test_evolution_2026_08_03.py`: 21 tests covering all changes

## Checks
- ruff check ✓
- ruff format ✓ (memory_tool, test file; tool_executor has pre-existing format issues)
- targeted tests ✓ (21 passed)

### DIFF
--- a/agent/tool_executor.py
+++ b/agent/tool_executor.py
@@ -95,9 +95,9 @@ def _parse_tool_arguments(
             {
                 "error": "Missing tool arguments",
                 "message": (
-                    "No arguments were provided. Check the tool's parameter "
-                    "schema and provide a valid JSON object with all required "
-                    "fields."
+                    "The tool was not executed. No arguments were provided. "
+                    "Check the tool's parameter schema and provide a valid "
+                    "JSON object with all required fields."
                 ),
             },
             ensure_ascii=False,
@@ -125,9 +125,10 @@ def _parse_tool_arguments(
             {
                 "error": "Invalid tool arguments (JSON parse error)",
                 "message": (
-                    f"Tool arguments could not be parsed as JSON: {exc.msg} "
-                    f"at position {exc.pos}. Provide a valid JSON object "
-                    f"with the correct field names and types for this tool."
+                    f"The tool was not executed. Its arguments could not be "
+                    f"parsed as JSON: {exc.msg} at position {exc.pos}. "
+                    f"Provide a valid JSON object with the correct field "
+                    f"names and types for this tool."
                 ),
             },
             ensure_ascii=False,
@@ -137,9 +138,10 @@ def _parse_tool_arguments(
             {
                 "error": "Invalid tool arguments (wrong type)",
                 "message": (
-                    f"Tool arguments must be a JSON string, got "
-                    f"{type(raw_arguments).__name__}. Provide a valid JSON "
-                    f"object with the correct field names for this tool."
+                    f"The tool was not executed. Its arguments must be a "
+                    f"JSON string, got {type(raw_arguments).__name__}. "
+                    f"Provide a valid JSON object with the correct field "
+                    f"names for this tool."
                 ),
             },
             ensure_ascii=False,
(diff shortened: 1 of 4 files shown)


Lexus2016:
Pushed a fix on top: **the sub-classification dropped a signal the old single message carried** — that the call was *rejected*, not *executed and failed*.

The model needs that distinction to decide whether re-issuing the call makes sense. None of the four new messages said it, so a malformed-args rejection now reads to the model like a tool that ran and errored.

All four now open with `The tool was not executed.` — missing arguments, JSON parse error, wrong argument type, parsed-but-not-an-object.

Also updated `test_malformed_tool_arguments` to assert the error *prefix* rather than the exact pre-#1647 string (sub-classification is the whole point of the change), and added an assertion on the not-executed signal so it cannot regress silently again.

467 tests pass across `test_malformed_tool_arguments`, `test_run_agent` and `test_evolution_2026_08_03`.

---

On #1648's "spiral cap" half: no code needed. `memory` is already in `_SPIRAL_PRONE_TOOLS`, and `_memory_enriched_error` returns an `error` key — which is the part that matters. The cross-turn `spiral_failure_cap` only counts results that `classify_tool_failure` sees as failures, so an enriched message *without* that key would have left the cap toothless. That exact gap was #1589 for `search_files`. This PR gets it right.

PR #1651, opened 2026-08-03
fix: terminal parse-error bare-string fallback + enriched error (#1647)

### LINKED ISSUE #1647: [FIX] terminal parse-error bucket at 285 failures/7d (rate 0.0184) — enrich malformed-arg error + bare-string fallback

Lexus2016:
Closes #1647

## Problem
The terminal tool's parse-error bucket is the second-largest terminal failure sub-bucket (285 failures/7d, rate 0.0184/session). Models frequently emit the command as a bare string (quoted '"ls -la"' or unquoted 'ls -la') instead of the schema's `{"command": ...}` object. These were rejected as parse errors with no indication of which field was malformed, so the agent could not self-correct and spiralled (up to 55 consecutive retries).

## Fix
In `agent/tool_executor.py` `_parse_tool_arguments`:
1. **Bare-string fallback** for the terminal tool: a quoted or unquoted command string maps onto the `command` field and executes instead of being rejected.
2. **Enriched parse-error message**: the error now names the tool and includes the underlying exception type (e.g. `JSONDecodeError`) so the agent can correct it on retry.

## Tests
`tests/run_agent/test_terminal_parse_error_1647.py` — 6 tests covering valid-object passthrough, quoted/unquoted bare strings, empty-string rejection, non-terminal unchanged, and error enrichment.

## Verification
`pytest tests/run_agent/test_run_agent.py::TestConcurrentToolExecution tests/run_agent/test_tool_arg_coercion.py tests/run_agent/test_terminal_parse_error_1647.py` — all pass.

Note: the spiral-prone tool cap already includes `terminal` (item 3 of #1647), so this PR covers items 1 and 2.

### DIFF
--- a/agent/tool_executor.py
+++ b/agent/tool_executor.py
@@ -76,19 +76,38 @@ _MAX_TOOL_WORKERS = 8
 _DEFAULT_CONCURRENT_TOOL_TIMEOUT_S = 420.0
 
 
-def _parse_tool_arguments(raw_arguments: Any) -> tuple[dict, Optional[str]]:
-    """Parse model-emitted arguments without repairing or coercing them."""
+def _parse_tool_arguments(
+    raw_arguments: Any, tool_name: str = ""
+) -> tuple[dict, Optional[str]]:
+    """Parse model-emitted arguments without repairing or coercing them.
+
+    Bare-string fallback for the terminal tool (#1647): models frequently emit
+    the command as a plain string (a quoted JSON string ``"ls -la"``, or an
+    unquoted ``ls -la`` that fails ``json.loads``) instead of the schema's
+    ``{"command": ...}`` object. Both shapes are mapped onto the ``command``
+    field so a well-formed command executes instead of being rejected as a
+    parse error — the terminal parse-error bucket was 285 failures/7d with no
+    way for the agent to self-correct.
+    """
+    parse_error = None
     try:
         arguments = json.loads(raw_arguments)
-    except (json.JSONDecodeError, TypeError):
+    except (json.JSONDecodeError, TypeError) as exc:
         arguments = None
+        parse_error = exc
     if isinstance(arguments, dict):
         return arguments, None
+    if tool_name == "terminal":
+        bare = arguments if isinstance(arguments, str) else raw_arguments
+        if isinstance(bare, str) and bare.strip():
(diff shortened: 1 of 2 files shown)


Lexus2016:
Closed as implemented_on_main: commit 42c68b49c (#1649) already implements this — 'feat: enrich parse-error messages + terminal bare-string fallback + memory error decomposition (Closes #1647, #1648)'. The merged version sub-classifies into four error cases and preserves the 'tool was not executed' signal, which is more complete than this PR. No additional change needed.

PR #1652, opened 2026-08-03
fix: decompose memory 'other' error into structured reasons (#1648)

### LINKED ISSUE #1648: [FIX] memory tool 'other' error bucket persists at 133 failures/7d (98% of memory failures) — decompose + spiral cap

Lexus2016:
Closes #1648

## Problem
The memory tool's 'other' error bucket is the dominant memory failure mode (133 failures/7d, 98% of memory failures). The opaque 'other' error gives the model no way to tell a transient connection issue from a permanent schema mismatch or capacity limit, driving blind retry spirals (up to 11 consecutive identical calls across 24 sessions).

## Fix
In `tools/memory_tool.py`:
1. **`_classify_memory_failure`** — decomposes an unexpected exception into a structured `reason` category aligned with the introspection taxonomy: `connection-timeout`, `serialization-error`, `schema-mismatch`, `capacity-exceeded`, or `other`.
2. **`_enrich_memory_error`** — returns a JSON error with the reason + a concrete recovery directive (retry-with-backoff vs restructure-payload vs consolidate vs skip).
3. **Dispatch wrap** — the `memory_tool` store dispatch now catches any escaping exception and enriches it, so the model gets actionable guidance instead of an opaque error.

## Tests
`tests/run_agent/test_memory_error_decomposition_1648.py` — 7 tests covering each category, the structured error shape, and the dispatch wrap.

## Verification
`pytest tests/tools/test_memory_tool.py tests/run_agent/test_memory_error_decomposition_1648.py` — all pass.

Note: the spiral-prone tool cap already includes `memory` (item 3 of #1648), so this PR covers items 1 and 2.

### DIFF
--- a/tests/run_agent/test_memory_error_decomposition_1648.py
+++ b/tests/run_agent/test_memory_error_decomposition_1648.py
@@ -1,0 +1,70 @@
+"""Tests for issue #1648 (memory 'other' error decomposition).
+
+The memory tool historically returned opaque "other" errors — the dominant
+memory failure bucket (98%) that drove blind retry spirals. This asserts that
+unexpected exceptions are decomposed into structured ``reason`` categories
+with recovery directives so the model can act instead of looping.
+"""
+
+import json
+
+from tools.memory_tool import (
+    _classify_memory_failure,
+    _enrich_memory_error,
+    memory_tool,
+)
+
+
+class TestMemoryFailureClassification:
+    def test_connection_timeout(self):
+        tag = _classify_memory_failure(TimeoutError("file lock timed out"))
+        assert tag["reason"] == "connection-timeout"
+        assert "retry" in tag["recovery"].lower()
+
+    def test_serialization_error(self):
+        tag = _classify_memory_failure(ValueError("cannot serialize JSON"))
+        assert tag["reason"] == "serialization-error"
+
+    def test_schema_mismatch(self):
+        tag = _classify_memory_failure(TypeError("invalid argument type"))
+        assert tag["reason"] == "schema-mismatch"
+
+    def test_capacity_exceeded(self):
+        tag = _classify_memory_failure(OSError("memory limit exceeded"))
+        assert tag["reason"] == "capacity-exceeded"
+        assert "consolidate" in tag["recovery"].lower()
+
+    def test_unknown_falls_back_to_other(self):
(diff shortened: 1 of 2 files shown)


Lexus2016:
Closed as implemented_on_main: commit 42c68b49c (#1649) already implements this — 'feat: enrich parse-error messages + terminal bare-string fallback + memory error decomposition (Closes #1647, #1648)'. The merged version decomposes the memory 'other' bucket on main. No additional change needed.

PR #1677, opened 2026-08-04
fix: stagger cron schedules to prevent API rate-limit collisions (Closes #1673)

### LINKED ISSUE #1673: [FIX] Provider rate limiting (zai/glm-5.2) — stagger cron schedules + concurrency limiter (~90 HTTP 429s)
## Evidence

**Pattern ID:** INTROS-004 | **Severity:** MEDIUM | **Priority score:** 0.80 | **Estimated effort:** 0.2

~90 HTTP 429 errors across Aug 2–4, concentrated when multiple cron jobs fire simultaneously. Two distinct error codes from the zai provider: **1302** (`Rate limit reached for requests`) and **1305** (`The service may be temporarily overloaded`). The existing retry policy (8 attempts, exponential backoff) handles most cases, but some calls fail after all retries, and the retries themselves waste API budget and add latency.

### Error counts (from introspection tick57)
- `rate_limit_error_http_429`: **~90 occurrences** across Aug 2–4
- 64 RateLimitError total in pass3 analysis introspection_findings
- Clusters at hour boundaries where multiple cron jobs overlap

### Log samples
```
2026-08-02 13:00:25 WARNING cron_bc6edf39bede API call failed error_type=RateLimitError HTTP 429: Rate limit reached for requests
```
- Two concurrent sessions (20260804_100049_ddc049 and cron_3ba3e5071b30_20260804) both hitting 429 at 10:01 on Aug 4, each on retry attempt 3/8

### Cron overlap (from jobs.json)
| Job | Schedule | Overlaps at |
|-----|----------|-------------|
| hydra (bc6edf39bede) | `*/30` | Every 30 min |
| research (3ba3e5071b30) | `0 9` | 09:00 |
| analysis (908e3d35d43d) | `0 1,5,9,13,17,21` | 01:00, 05:00, **09:00**, 13:00, 17:00, 21:00 |

Multiple jobs overlap at several hour boundaries (notably 09:00 where hydra + research + analysis all fire).

### Prior tracking
- No FIX issue filed for rate-limit staggering (analysis confirms status `no-issue-filed`)

## Root cause
No global coordination between cron jobs — they fire on wall-clock boundaries and all hit the zai/glm-5.2 API within the same few seconds, exceeding the rate limit. The retry backoff eventually clears most requests, but at the cost of wasted budget and latency.

## Acceptance criteria (≤200 lines)
- [ ] **Stagger cron schedules**: add 30–60s jitter offsets (or fixed offsets) so concurrent jobs do not all fire at the same wall-clock second. E.g., hydra at `*/30` +0s, analysis at `0 1,5,9...` +30s offset, research at `0 9` +60s offset.
- [ ] **OR** implement a lightweight global concurrency limiter for API calls across parallel cron sessions (e.g., a file-based or semaphore-based lock that caps concurrent provider calls to N). Pick whichever is simpler within the 200-line cap — staggering schedules is the lower-effort option.
- [ ] After the fix, the number of HTTP 429 errors in a 24h window should drop significantly (target: <50% of current ~90/day).
- [ ] Do NOT change the retry policy itself (8 attempts, exponential backoff is fine) — the goal is to prevent the collisions, not to mask them with more retries.
- [ ] Verify: run a full day of cron ticks and confirm no concurrent 429 clusters at hour boundaries.

## Implementation notes
- **Lowest-effort option:** add fixed offsets to the cron schedule expressions (0 lines of code, just config changes). This alone may resolve the majority of collisions.
- The concurrency limiter is the more robust option but higher effort — only do it if staggering alone is insufficient. Keep within 200 lines.

Lexus2016:
## Summary

Stagger evolution cron schedules to prevent concurrent API calls from triggering HTTP 429 rate-limit errors (~90/day from zai/glm-5.2).

## Fix (config-only, 3 schedule changes)
- analysis: `0 1,5,9,...` → `5 1,5,9,...` (+5 min)
- research: `0 9` → `15 9` (+15 min)
- ci-diagnosis: `*/30` → `10,40` (+10 min)

## Tests
Added regression test ensuring schedules stay staggered. 86 tests pass.

Closes #1673

### DIFF
--- a/cron/evolution/analysis.yaml
+++ b/cron/evolution/analysis.yaml
@@ -1,5 +1,5 @@
 name: evolution-analysis
-schedule: "0 1,5,9,13,17,21 * * *"  # Every 4h (was daily 21:00). Raises processing throughput vs ~25 issues/day generation. 21:00 slot still follows introspection (20:00). Watchdog STAGES mirrors the FIRST slot (1).
+schedule: "5 1,5,9,13,17,21 * * *"  # #1673: staggered +5 min off :00 to avoid API rate-limit collisions with hydra/research/ci-diagnosis firing at the same wall-clock second.
 enabled: true
 mode: PRIVATE
 
--- a/cron/evolution/ci-diagnosis.yaml
+++ b/cron/evolution/ci-diagnosis.yaml
@@ -2,7 +2,7 @@ name: evolution-ci-diagnosis
 # Polls open evolution PRs for CI failures and either auto-fixes trivial ones
 # or creates focused child issues for complex failures.
 # Runs every 30 minutes; bounded at 48 runs/day by the scheduler.
-schedule: "*/30 * * * *"
+schedule: "10,40 * * * *"  # #1673: staggered off :00/:30 to avoid colliding with hydra (*/30) and LLM stages at :00
 enabled: true
 mode: PRIVATE
 
--- a/cron/evolution/research.yaml
+++ b/cron/evolution/research.yaml
@@ -1,5 +1,5 @@
 name: evolution-research
-schedule: "0 9 * * *"  # Daily at 9 AM
+schedule: "15 9 * * *"  # #1673: staggered +15 min off :00 to avoid collision with hydra (*/30) and analysis (:05) at the 09:00 boundary.
 enabled: true
 mode: PUBLIC
 
--- a/tests/scripts/test_evolution_cron_stagger.py
+++ b/tests/scripts/test_evolution_cron_stagger.py
@@ -1,0 +1,92 @@
+"""Regression test for issue #1673 — cron schedule staggering.
+
+Ensures evolution cron jobs do NOT all fire at the same wall-clock minute
+(:00), which caused ~90 HTTP 429 errors from concurrent API calls to the
+zai/glm-5.2 provider.
+
+The test checks that the highest-frequency LLM-consuming jobs (analysis,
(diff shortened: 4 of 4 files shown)


Lexus2016:
Superseded by #1678 (merged) plus #1679.

**Why #1678 was taken instead:** it staggers all eight LLM-hitting pipeline
stages onto distinct minute offsets; this PR moved three. The two overlap on
`analysis.yaml` and `research.yaml`, so they could not both land.

**Your regression test's idea was the right one and is not lost** — #1679 carries
it forward, with one change worth stating plainly. The per-job form here
(`assert minute != "0"` for analysis / research / ci-diagnosis) would have passed
against the schedules #1678 actually shipped, while three pairs were still
colliding:

| pair | collides at |
|---|---|
| `ci-diagnosis` × `postmortem-miner` | 06:00 |
| `ci-diagnosis` × `research` | 09:30 |
| `funnel` × `integration` | 07:40 |

`ci-diagnosis` kept `*/30` (fires :00 and :30), and `test_ci_diagnosis_off_zero_minute`
here only checks the literal string `*/30 * * * *` — so a rewrite to `0,30` would
pass while colliding identically.

#1679 asserts the invariant instead: expand every schedule to its full
`(hour, minute)` firing set, intersect every pair, fail on any shared minute. It
holds when schedules legitimately move, and it names the offending pair.

Thanks — this PR is what surfaced that #1678 needed a guard at all.

PR #1705, opened 2026-08-05
fix: patch replace-mode + empty old_string non-retryable diagnostic (Closes #1703)

### LINKED ISSUE #1703: [FIX] patch tool replace-mode + empty old_string produces 5-8 retry spiral — non-retryable diagnostic needed
The agent repeatedly invokes the patch tool in replace mode with an empty `old_string`, receiving an error each time, then retries 5-8 times before recovering. This wastes tool calls and stalls the edit.

- Frequency: 12 occurrences of the patch-old-string-empty reason in the last 7 days (2215 sessions scanned); 5 sessions hit patch spirals up to 8 consecutive retries
- Tool/area involved: patch tool (replace mode)
- Failure shape: agent calls patch with mode=replace but passes an empty `old_string` (or omits it). The tool returns an error like "old_string is empty / required for replace mode" — the agent retries the same shape 5-8 times before either switching to patch mode or giving up.
- Rate trend: 0.0054 failures/session this window; stable vs prior windows (no prior baseline recorded)

### Impact on real tasks
Each patch-old-string-empty failure costs 5-8 wasted tool calls and stalls the edit until the agent either gives up or stumbles onto the correct mode. In 5 sessions this produced a patch spiral that consumed significant context budget and delayed the actual file edit.

### Proposed direction
- Detect the replace-mode + empty-old_string shape at the tool boundary and return a single, actionable, non-retryable diagnostic: explicitly state "replace mode requires a non-empty old_string; use mode=patch for insertions or supply the text to replace."
- Treat the second identical failure of this shape as non-retryable (loop-guard style) to break the spiral at 2 instead of 8.
- Optional: when mode is omitted and old_string is empty, auto-suggest mode=patch.

### Value
- Impact: 0.5 (moderate — wastes tool budget in 5-8 call spirals, but bounded; blocks the edit until recovery)
- Effort: 0.2 (clear validation branch in the patch tool error path; small)
- Priority Score: 0.92

Lexus2016:
## Summary

Fixes the patch tool replace-mode + empty `old_string` retry spiral (#1703).

**Problem:** The agent calls `patch` with `mode=replace` but passes an empty `old_string=""`, receives an error, and retries 5-8 times before recovering. The root cause: the error message didn't mention `mode=patch` as an alternative, and the consecutive-failure tracker only fires for `no_match`/`ambiguous` errors — never for `old_string_empty`.

**Evidence (introspection):** 12 failures/7d, 5 sessions with spirals up to 8 consecutive retries. Rate: 0.0054 failures/session.

## Changes

| File | Change |
|------|--------|
| `tools/file_tools.py` | Early gate for `old_string==""` before fuzzy_match. Records failure via `_record_patch_failure`, escalates to `NON-RETRYABLE` after 2 consecutive. Suggests `mode='patch'` for insertions. |
| `tools/file_operations.py` | Updated `old_string_empty` recovery hints (both inline classify + structured subclass) to mention `mode='patch'`. |
| `tools/fuzzy_match.py` | Updated `old_string_empty` structured recovery to suggest `mode='patch'`. |
| `tests/tools/test_patch_empty_old_string_1703.py` | 5 new tests: first-occurrence hint, read_file direction, NON-RETRYABLE escalation, None vs empty string, valid patch not intercepted. |
| `tests/tools/test_file_error_classification.py` | 2 updated assertions: `old_string_empty` recovery must contain `mode='patch'`. |

## How it breaks the spiral

1. **1st empty attempt:** Actionable error with `mode='patch'` suggestion → agent learns the alternative
2. **2nd consecutive empty attempt:** `NON-RETRYABLE` escalation → explicitly tells agent to stop retrying

## Checks
- ✅ Lint: `ruff check` clean
- ✅ Format: `ruff format --check` clean (on changed files)
- ✅ Tests: 31 passed (5 new + 26 existing classification tests)
- ✅ Diff size: 62 lines (well under 200-line merge cap)

Closes #1703

### DIFF
--- a/tests/tools/test_file_error_classification.py
+++ b/tests/tools/test_file_error_classification.py
@@ -13,7 +13,9 @@ class TestClassifyFileError:
         assert classify_file_error("") is None
 
     def test_permission(self):
-        klass, rec = classify_file_error("Write denied: '/etc/x' is a protected system/credential file.")
+        klass, rec = classify_file_error(
+            "Write denied: '/etc/x' is a protected system/credential file."
+        )
         assert klass == "permission" and "allowed path" in rec
 
     def test_not_found_without_similars_routes_to_write_file(self):
@@ -22,7 +24,8 @@ class TestClassifyFileError:
 
     def test_not_found_with_similars_is_fuzzy_match(self):
         klass, rec = classify_file_error(
-            "File not found: /tmp/utils.py", similar_files=["/tmp/util.py", "/tmp/utils2.py"]
+            "File not found: /tmp/utils.py",
+            similar_files=["/tmp/util.py", "/tmp/utils2.py"],
         )
         assert klass == "fuzzy_match" and "util.py" in rec
 
@@ -57,7 +60,9 @@ class TestClassifyFileError:
         assert "replace_all" in rec or "context" in rec
 
     def test_verification(self):
-        klass, _ = classify_file_error("Post-write verification failed: could not re-read x")
+        klass, _ = classify_file_error(
+            "Post-write verification failed: could not re-read x"
+        )
         assert klass == "verification"
 
     def test_binary(self):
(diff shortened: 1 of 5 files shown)


Lexus2016:
Closing: superseded by PR #1710 (the newer green successor for the same issue #1703), which has been merged. This branch was also red on CI (3 failing checks) and #1703 is now closed.

Issue #1744, opened 2026-08-06
[FIX] patch no-match rate 0.0468 — disambiguation fix did not reduce the no-match case


Lexus2016:
## Problem (observed in real usage)
The patch tool's no-match failure remains the dominant patch error mode despite a closed fix that proposed returning candidate line numbers for disambiguation.

### Evidence (aggregated, anonymized)
- Frequency: patch-no-match at rate **0.0468 per session** (113 failures / 2414 sessions, 7-day window)
- This is **61%** of all patch failures (113 of 185)
- Tool/area involved: patch tool, fuzzy-match no-match reason bucket
- Failure shape: agent supplies an old_string that the fuzzy matcher cannot locate; the agent re-reads the file and retries with a near-identical string

### Impact on real tasks
Each no-match failure stalls an in-progress edit, forcing a re-read + retry cycle (up to 8 consecutive in 5 sessions). The prior fix (#1683) proposed returning candidate match locations but the no-match rate is unchanged at scale — the disambiguation cue is either not implemented or not effective for the no-match case (vs. the ambiguous case which is now low at 0.0025).

### Proposed direction
1. For no-match: return the **nearest lines** to the supplied anchor (Levenshtein-nearest) + file length, so path/typo drift is visible in one step instead of a blind re-read
2. Add a strict (non-fuzzy) first pass when the anchor is quoted — quoted anchors that no-match indicate content drift the fuzzy layer masks
3. After N=3 consecutive patch failures in a session, inject a directive to use write_file (full replace) instead of continuing to retry patch

### Value
- Impact: 0.7 (frequent, blocks in-progress edits, recoverable but wasteful)
- Effort: 0.3
- Priority Score: 0.7 × 2 × (1 − 0.4 × 0.3) = **1.144**

Issue #1683, opened 2026-08-04
[FIX] patch ambiguous/no-match fuzzy-match failures recur at scale — 371/7d despite line-content enrichment


Lexus2016:
## Problem (observed in real usage)
The agent repeatedly fails to apply targeted file edits because the patch tool cannot resolve an unambiguous match location.

### Evidence (aggregated, anonymized)
- Frequency: the patch tool accounted for 528 failures in the last 7 days across 16,991 scanned sessions.
- Tool/area involved: patch tool (fuzzy-match mode).
- Failure shape: "patch-ambiguous" (200 events) + "patch-no-match" (171 events) = 371 of 528 failures (~70%). These two buckets alone exceed the volume of every other patch error class (parse-error 40, file-not-found 21, other 82).
- Rate: patch failure rate 0.031/session is the second-highest of any non-terminal tool.

### Impact on real tasks
Every ambiguous/no-match failure stalls an edit the agent is actively performing, forcing a re-read of the file and a retry. At this volume it is the single largest source of edit-loop friction that has not been resolved by the earlier line-content enrichment (#976) or the non-retryable diagnostics (#1612). The agent is left re-guessing an insertion point instead of getting a deterministic disambiguation cue.

### Proposed direction
Make the ambiguous/no-match error actionable at the point of failure: when a patch target string matches more than one location, return the line numbers + surrounding context of each candidate match (not just "ambiguous") so the model can pick a unique anchor in one step; when no match is found, return the nearest lines to the supplied anchor and the file length so a path/typo drift is visible instead of silent. Consider a strict (non-fuzzy) first pass when the anchor is quoted, since quoted anchors that still no-match indicate content drift the fuzzy layer masks.

### Value
- Impact: 0.6 (frequent, blocks in-progress edits, but recoverable)
- Effort: 0.3
- Priority Score: 1.06

Lexus2016:
[...]

Issue #1681, opened 2026-08-04
[FIX] read_file receives stringified-array and directory paths after coerce_tool_args drops malformed list params — file-not-found spiral


Lexus2016:
## Problem (observed in real usage)
When the model passes a stringified JSON array (e.g. `"[\"path/a\", \"path/b\"]"`) or a directory path to `read_file`'s `path` parameter, the `coerce_tool_args` layer fails to parse it and silently drops the argument. `read_file` then receives a malformed path (the literal stringified array or a directory), returns "File not found", and the agent retries without changing strategy — a spiral.

### Evidence (aggregated, anonymized)
- Frequency: ~15 occurrences in the most recent 2-day window; recurring across multiple sessions
- Tool/area involved: `read_file` + `coerce_tool_args` interplay
- Failure shape: model emits a list of paths for a parameter that expects a single path (or a bare path list for a batch-read parameter); coercion drops it; read_file gets `[/some/path, /other/path]` as a literal string and returns file-not-found; a related sub-pattern is passing a directory path where a file is expected

### Impact on real tasks
This directly feeds the highest-volume core-tool failure signal: `read_file:file-not-found` sits at ~7.1% failure rate (143 failures over 2008 sessions/7d), a large share of which traces to this coercion-to-malformed-path path rather than genuine missing files. Each occurrence wastes 2–6 API turns as the agent blind-retries the same path before correcting, and in cron/unattended contexts there is no human to notice the wrong path.

### Proposed direction
When `coerce_tool_args` cannot parse a list-typed argument, detect whether the raw value is a bracket-wrapped or comma-separated list of paths and split accordingly before dropping. Additionally, add a "path is a directory, use search_files instead" hint in the `read_file` file-not-found error when the path resolves to a directory, so the agent gets actionable feedback instead of a generic error.

### Value
- Impact: 0.6 (high-volume signal, feeds the #3 core-tool failure rate, blocks autonomous recovery)
- Effort: 0.3 (targeted coercion-path fix + error hint)
- Priority Score: 0.6 × 2 × (1 − 0.4 × 0.3) = **1.05**

Lexus2016:
Re-opened by realized-impact gate: post-merge verdict for PR #1684 was **regressed** — read_file file-not-found rate 0.071→0.0746 post-merge. The stringified-array coercion fix did not reduce the target signal. Needs a focused regression before re-closing.

PR #1749, opened 2026-08-06
fix: syntax-error classification, patch no-match file length, comma-separated read_file path coercion (#1743 #1744 #1681)

### LINKED ISSUE #1743: [FIX] Terminal non-zero-exit spiral regressed (4th recurrence) — 0.1442 rate, 33-deep, 213 sessions
The terminal tool's non-zero-exit failure rate is the highest of any core tool and its retry spiral depth exceeds the success criteria set by the most recent fix.

- Frequency: non-zero-exit terminal failures occur at rate **0.1442 per session** (348 failures / 2414 sessions scanned, 7-day window)
- Spiral depth: up to **33 consecutive** terminal calls in a single session, across **213 sessions**
- Tool/area involved: terminal tool, non-zero-exit reason bucket
- Failure shape: shell commands exit non-zero; the agent retries (often with near-identical commands) instead of switching approach

### Impact on real tasks
Each non-zero-exit spiral wastes 10-30+ tool calls and stalls the user's task. The prior fix (#1611) set explicit success criteria: <5 consecutive across <50 sessions. Current state is **33-deep across 213 sessions** — the criteria are not met. The circuit-breaker infrastructure exists but is not covering all non-zero-exit subclasses.

### Proposed direction
1. Audit which non-zero-exit subclasses bypass the existing circuit breaker (exit codes 1, 126, 127, timeout, signal-kill)
2. Make non-retryable diagnostics terminal: when the exit code indicates a permanent failure (command not found=127, permission denied=126, syntax error), return a non-retryable diagnostic that names the error class and suggests the alternative action
3. After N=3 consecutive non-zero exits of the same tool in a session, inject a fallback directive directing the agent to an alternative approach

### Value
- Impact: 0.9 (highest-volume failure signal, blocks shell-based tasks in 213 sessions)
- Effort: 0.4 (circuit breaker exists, needs broader coverage + non-retryable classification)
- Priority Score: 0.9 × 2 × (1 − 0.4 × 0.4) = **1.512**

Lexus2016:
Closes #1743, #1744, #1681 (all three introspection-selected, genuine real-usage bug fixes).

## #1743 — Terminal non-zero-exit spiral (syntax-error gap)
The circuit breaker classified 127/126/timeout/signal-kill as non-retryable, but **syntax errors** (exit code 2, or `syntax error` / `unexpected token` output) fell through to generic `persistent_error`, so the agent got no targeted "this is a syntax error, re-running won't help" cue.

- Add `_SYNTAX_ERROR_PATTERNS`; classify as deterministic non-retryable with a "fix the syntax / use execute_code / do NOT retry" hint.
- **Not** gated on bare `exit_code == 2` — grep/tar/diff use exit-2 for generic errors (file-not-found, archive corruption) that aren't syntax. Fires only on syntax-error text OR exit-2 from a known interpreter (`bash`/`python`/`node`/…).

## #1744 — Patch no-match rate (file-scale cue)
The nearest-lines disambiguation shipped in #1683 was already present (that's why no-match was unchanged). The remaining gap: the no-match error gave no sense of **file scale**, so the model couldn't decide re-read vs write_file in one step.

- `format_structured_error` no-match now prepends `File length: N lines, M chars` before the closest-matching region. A tiny file → cheap full re-read; a large one → write_file/search.

## #1681 — read_file stringified/directory paths (comma-list coercion)
`_extract_first_from_stringified_array` handled JSON arrays (`'["a","b"]'`) but NOT non-JSON bracket/bare comma lists (`[path/a, path/b]` or `src/a.py, src/b.py`) — those fell through `json.loads` → `None` → dropped → read_file file-not-found spiral.

- Add a comma-separated fallback that extracts the first path-like token.
- Widen the string-typed trigger to also fire on comma-containing strings.
- Path-like heuristics (require a path separator in both tokens) prevent mangling sentences/bare words.

## Verification
- 155 tests pass across the 5 touched/related test files.
- New tests: +37 classifier, +28 fuzzy_match, +34 coercion.

### DIFF
--- a/model_tools.py
+++ b/model_tools.py
@@ -874,7 +874,14 @@ def coerce_tool_args(tool_name: str, args: Dict[str, Any]) -> Dict[str, Any]:
         if (
             expected == "string"
             and isinstance(value, str)
-            and value.lstrip().startswith("[")
+            and (
+                value.lstrip().startswith("[")
+                # #1681 — also fire on bare comma-separated path lists
+                # (``"src/a.py, src/b.py"``, no brackets). The helper's
+                # path-like guards reject non-path strings (sentences,
+                # bare words), so widening the trigger is safe.
+                or "," in value
+            )
         ):
             extracted = _extract_first_from_stringified_array(value)
             if extracted is not None:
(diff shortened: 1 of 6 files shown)


Lexus2016:
Superseded by PR #1750 (#1743 slice) and PR #1751 (#1744 + #1681 slice) — this oversized PR was split into two ≤200-line mergeable slices per the landability gate.

PR #1750, opened 2026-08-06
fix: classify terminal syntax errors as deterministic non-retryable (#1743)

### LINKED ISSUE #1743: [FIX] Terminal non-zero-exit spiral regressed (4th recurrence) — 0.1442 rate, 33-deep, 213 sessions

Lexus2016:
Closes #1743 (slice 1).

## Problem
The terminal circuit breaker classified exit 127 (missing_command), 126 (permission_denied), timeouts, and signal-kill as non-retryable — but **syntax errors** (exit code 2, or 'syntax error'/'unexpected token' output) fell through to generic persistent_error. The agent got no targeted cue that the command itself is malformed and re-running won't help, feeding the non-zero-exit retry spiral.

## Fix
- Add `_SYNTAX_ERROR_PATTERNS` (`syntax error`, `unexpected token`, `unexpected end of file`, `SyntaxError:`).
- Classify as deterministic non-retryable persistent_error with a targeted 'fix the syntax / use execute_code / do NOT retry the same command' hint.
- **Not** gated on bare `exit_code == 2` — grep/tar/diff use exit-2 for generic errors (file-not-found, archive corruption) that aren't syntax. Fires only on syntax-error text OR exit-2 from a known interpreter (bash/sh/python/node/ruby/perl/php).

## Verification
+37 lines of new classifier tests; all 155 related tests pass. Slice is 75 changed lines, under the 200-line self-merge cap.

### DIFF
--- a/tests/tools/test_terminal_failure_classifier.py
+++ b/tests/tools/test_terminal_failure_classifier.py
@@ -117,6 +117,43 @@ class TestClassifyPersistentError:
         assert result.should_retry is False
 
 
+class TestClassifySyntaxError:
+    """Syntax errors are deterministic and must not be blind-retried (#1743)."""
+
+    def test_syntax_error_text_pattern(self):
+        """Bash 'syntax error near unexpected token' is non-retryable."""
+        result = classifier.classify_terminal_failure(
+            "if then", 2, "", "bash: syntax error near unexpected token"
+        )
+        assert result.category == classifier.FailureCategory.persistent_error
+        assert result.should_retry is False
+        assert "Syntax" in result.hint
+
+    def test_interpreter_exit_2_is_syntax(self):
+        """Python exit code 2 (interpreter) is treated as a syntax error."""
+        result = classifier.classify_terminal_failure(
+            'python3 -c "if true print(1)"', 2, "", ""
+        )
+        assert result.category == classifier.FailureCategory.persistent_error
+        assert result.should_retry is False
+        assert "Syntax" in result.hint
+
+    def test_grep_exit_2_not_syntax(self):
+        """grep exit 2 is a generic error (file-not-found), NOT a syntax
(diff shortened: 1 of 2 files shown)


Lexus2016:

PR #1791, opened 2026-08-07
feat: auto-dispatch core tools routed through tool_call bridge (Closes #1786)

### LINKED ISSUE #1786: [FIX] tool_call bridge mis-routing — 329 failures/273 sessions, model dispatches core tools through the bridge
The agent repeatedly mis-routes directly-available tools through the `tool_call` bridge, which rejects them with "is not a deferrable tool." This wastes turns and sometimes causes retry spirals when the model does not understand the error.

- Frequency: 329 failures across 273 of ~21,700 recent sessions (failure rate 0.0133/session)
- Tool/area: `tool_call` bridge dispatcher
- Failure shape breakdown:
  - 61% (201): "[TOOL_NAME] is not a deferrable tool" — model calls core tools (execute_code, terminal, search_files, memory, etc.) via the bridge instead of directly
  - 16% (53): "arguments is not valid JSON" — model emits malformed JSON in the arguments field
  - 14% (47): "tool_call cannot invoke tool_call (bridge self-nesting)" — model nests the bridge inside itself
  - 7% (23): "requires a name argument" — model omits the required name field
- Spiral data: max 13 consecutive identical runs, 50 sessions affected

### Impact on real tasks
Every mis-routed call costs a full round-trip. When the model retries the same mis-routed call (up to 13 times), the session stalls. The dominant sub-pattern (61%) means the model fundamentally does not understand which tools are "deferrable" vs "core" — a systematic dispatch misunderstanding that repeats in every session where MCP tools are present.

### Proposed direction
1. Improve the `tool_call` error message to explicitly name the tool and say "call [TOOL_NAME] directly — it is already in your tool list." The current message is generic.
2. Add a system-prompt clarification distinguishing "always-available" tools from "deferred/on-demand" tools that require the bridge.
3. Consider auto-forwarding: if the named tool IS directly available, execute it directly instead of erroring (transparent dispatch fixup).

### Value
- Impact: 0.8 (273 sessions, systematic misunderstanding blocks tool dispatch)
- Effort: 0.4 (error message + prompt clarification)
- Priority Score: 0.8 × 2 × (1 − 0.4 × 0.4) = 1.34

Lexus2016:
Automated evolution PR for issue #1786.

## Problem
The model repeatedly mis-routes directly-available core tools (read_file, search_files, terminal, etc.) through the `tool_call` bridge, which rejects them with "is not a deferrable tool." This wastes turns and causes retry spirals — 329 failures across 273 sessions. The existing error message told the model to "call it directly" but the model ignored the guidance.

## Fix
When `resolve_underlying_call` detects a non-deferrable tool that IS in the effective core set, return it as dispatchable instead of returning an error. Both dispatch paths now auto-dispatch core tools:

1. **`resolve_underlying_call`** (tool_search.py): bypass the deferrability gate for effective core tools
2. **`model_tools.py` handle_function_call**: check `scoped_available_names` (includes core tools) in addition to `scoped_deferrable_names`
3. **`tool_executor.py` unwrap**: same dual-scope check for the concurrent execution path

Restricted sessions still reject out-of-scope tools — those tools never appear in `tool_defs`, so `scoped_available_names` excludes them.

## Tests
Updated 4 tests to reflect the new auto-dispatch behavior. All 87 tests in `test_non_deferrable_error.py` + `test_tool_search.py` pass.

## Impact
- 329 failures/273 sessions eliminated
- No behavior change for genuinely unavailable tools (still rejected with enriched error)
- No behavior change for valid deferred tool calls
- 198 changed lines (within 200-line self-merge cap)

### DIFF
--- a/agent/tool_executor.py
+++ b/agent/tool_executor.py
@@ -361,6 +361,44 @@ def _tool_search_scoped_names(agent) -> frozenset:
     return names
 
 
+def _tool_search_available_names(agent) -> frozenset[str]:
+    """Return ALL tool names in the session's tool_defs, including core tools
+    (#1786).  Used by the ``tool_call`` unwrap to auto-dispatch core tools."""
+    try:
+        import model_tools
+        from tools import tool_search as _ts
+        from tools.registry import registry as _registry
+    except Exception:
+        return frozenset()
+
+    enabled = getattr(agent, "enabled_toolsets", None)
+    disabled = getattr(agent, "disabled_toolsets", None)
+    cache_key = (
+        getattr(_registry, "_generation", 0),
+        frozenset(enabled) if enabled is not None else None,
+        frozenset(disabled) if disabled is not None else None,
+    )
+    cache_attr = "_tool_search_available_cache"
+    cached = getattr(agent, cache_attr, None)
+    if cached is not None and cached[0] == cache_key:
+        return cached[1]
+    try:
+        scoped_defs = model_tools.get_tool_definitions(
+            enabled_toolsets=enabled,
+            disabled_toolsets=disabled,
+            quiet_mode=True,
+            skip_tool_search_assembly=True,
+        ) or []
+        names = _ts.scoped_available_names(scoped_defs)
+    except Exception:
+        names = frozenset()
+    try:
+        setattr(agent, cache_attr, (cache_key, names))
+    except Exception:
+        pass
+    return names
+
+
 @dataclass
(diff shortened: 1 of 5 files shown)


Lexus2016:
Closing duplicate — PR #1793 is the correct, simpler fix for #1786 (enriched error message + system-prompt clarification). This PR's auto-dispatch approach in tool_executor.py is over-engineered for the issue scope.

PR #1818, opened 2026-08-08
feat: SkillTrojan defense — multi-skill composition trace (#1802)

### LINKED ISSUE #1802: [SLICE B] SkillTrojan defense: multi-skill composition trace (parent #1658)
## Problem

SkillTrojan distributes malicious payloads as fragments across multiple benign skills. Individually, each skill is harmless. Only when composed in a workflow does the payload reconstruct and execute. Static per-skill scanning (Slice A) cannot detect this.

## This Slice (B of B)

Implement **runtime composition tracing**: when multiple skills execute in a single workflow, trace the combined output for patterns indicating payload reconstruction.

### Implementation

1. **Composition context**: When 2+ skills execute in the same turn, activate a composition tracer that monitors combined output.
2. **Reconstruction detection**: Heuristics for payload assembly patterns:
   - Concatenation of encoded fragments from different skill outputs
   - Conditional execution triggered by cross-skill state
   - External resource access patterns that only emerge from composition
3. **Block + alert**: When reconstruction is detected, halt execution, alert the user with the implicated skills and detected pattern.

### Scope Guardrails (max 200 lines)

- Runtime heuristic only — no ML model.
- Depends on Slice A (#1801) for the scan infrastructure.

### Acceptance Criteria

- [ ] Multi-skill workflows emit a composition trace.
- [ ] Payload reconstruction patterns are detected and blocked.
- [ ] Alert includes implicated skills and detected pattern.
- [ ] Diff fits the 200-line self-merge cap.

### Estimated Effort

~100-150 lines. Effort: Medium (0.5). Depends on #1801.

## Source
- Parent: #1658
- SkillTrojan: ICML 2026, arXiv:2604.06811

Lexus2016:
Automated evolution PR for issue #1802. Runtime composition tracer that monitors combined output when 2+ skills execute in the same turn, detecting payload reconstruction patterns (base64 fragment assembly, cross-skill conditional triggers, URL assembly from fragments).

Closes #1802

### DIFF
--- a/tests/tools/test_skill_composition_tracer.py
+++ b/tests/tools/test_skill_composition_tracer.py
@@ -1,0 +1,225 @@
+"""Tests for tools/skill_composition_tracer.py — SkillTrojan composition defense."""
+
+import base64
+
+import pytest
+
+from tools.skill_composition_tracer import (
+    CompositionTracer,
+    check_composition,
+    reset_tracer,
+    ACTIVATION_THRESHOLD,
+)
+
+
+# ── Single skill: below threshold → no tracing ──────────────────────────
+
+
+class TestSingleSkillNoTrace:
+    def test_single_skill_returns_safe(self):
+        """A single skill should never trigger composition detection."""
+        reset_tracer()
+        result = check_composition("my-skill", "Some benign skill content here.")
+        assert result["success"] is True
+
+    def test_single_skill_with_suspicious_content(self):
+        """Even with base64 fragments, a single skill is below threshold."""
+        reset_tracer()
+        # A long base64 blob, but only one skill loaded.
+        blob = base64.b64encode(b"import os; os.system('id')").decode()
+        result = check_composition("lone-skill", blob)
+        assert result["success"] is True
+
+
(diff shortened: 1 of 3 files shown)


Lexus2016:
Blocked by code-review: regression. The composition tracer (`tools/skill_composition_tracer.py`) uses a module-level singleton (`_tracer = CompositionTracer()`) that accumulates skills across turns, but `reset_tracer()` is NEVER called in production code (only in tests). This violates the design invariant "2+ skills in the SAME turn" — in practice it becomes "2+ skills ever loaded in this process," causing spurious cross-turn blocks on legitimate skills loaded in later turns. This makes the agent WORSE after merging (anti-regression violation).

Fix: wire `reset_tracer()` into the turn boundary (e.g. at the top of the agent loop, or via an `on_turn_complete` hook) so the singleton resets between turns. Also tighten the cross-skill conditional heuristic — it currently fires on plain prose mentioning another skill (e.g. "after skill_X completed"), which will hard-block skills whose documentation mentions other skills.

Sending back for rework — see the comment on issue #1802.

Issue #1490, opened 2026-07-28
[FIX] write_file parse-error — 12 silent write failures/7d, zero existing coverage for this tool


Lexus2016:
## Problem (observed in real usage)
The file-writing tool fails 12 times with a parse-error (malformed payload that
fails input validation), plus 9 "other" and 3 "file-not-found" for 24 total
failures in the window. No issue in the entire tracker covers write_file — it is
a total coverage gap. A parse-error on write_file is especially costly because it
silently fails a file the user explicitly asked to create, and the agent may not
recognize the write did not land.

### Evidence (aggregated, anonymized)
- Frequency: 12 parse-error + 9 other + 3 file-not-found = 24 events (7-day window)
- write_file failure rate: 0.3% (low, but the parse-error path is unhandled)
- Also a contributor to the cross-tool parse-error cluster (484 events across 8 tools)

### Impact on real tasks
A silent failed write means the user's requested file is not created; the agent
proceeds as if it succeeded, producing downstream errors when the missing file is
later read.

### Proposed direction
Improve write_file input validation with a clear, actionable error on malformed
payloads (what field is wrong, what shape is expected) so the agent corrects and
re-emits rather than silently losing the write.

### Value
- Impact: 0.55
- Effort: 0.35
- Priority Score: 0.946

Lexus2016:
Rejected by evolution-analysis: duplicate — The write_file parse-error issue is a symptom of the systemic tool-argument parse-error problem across 8 tools (#1487), not a write_file-specific gap. The write_file tool already has a fail-closed pre-write syntax gate (file_operations.py:1575-1611) that validates JSON/YAML/TOML content before writing and returns a clear 'Refusing to write: candidate content fails syntax validation' error. The 12 silent write failures are from the agent emitting malformed tool-call arguments (the model's JSON generation layer), which is the same root cause as #1487's 484 events across 8 tools. Fixing write_file in isolation would not address the systemic argument-generation problem.

PR #1845, opened 2026-08-08
fix: add write_file to spiral-prone tools + enrich parse-error as non-retryable (Closes #1840)

### LINKED ISSUE #1840: [FIX] write_file parse-error spirals — 48 failures/7d (4× regression of #1490), 15-deep across 11 sessions
The write_file tool produces parse-error failures that cause retry spirals. A prior issue (#1490, closed 2026-07-31) reported 12 write_file parse-errors/7d — the current window shows 48 parse-errors (4× increase) with 15-deep consecutive spirals across 11 sessions. The original fix did not hold or a new parse-error mode emerged.

- Frequency: 48 parse-error failures across 11 of 23,547 recent sessions (failure rate 0.0020/session)
- Max consecutive retries: 15
- Tool/area involved: write_file tool, parse-error reason bucket
- Failure shape: write_file receives malformed content argument (likely JSON-encoded or incorrectly escaped payloads from coerce_tool_args) that fails content validation, triggering repeated retries with the same malformed input
- Prior fix: #1490 covered write_file parse-error at 12 failures/7d — now at 48, a 4× regression

### Impact on real tasks
When write_file fails with a parse-error, the agent cannot create or update files — a core capability. It retries the same malformed write up to 15 consecutive times, wasting turns. The agent does not recognize the parse-error as non-retryable and does not fall back to an alternative approach (e.g., terminal heredoc, reformatting the content argument).

### Proposed direction
1. Enrich the write_file parse-error to include the specific validation failure reason and a non-retryable diagnostic when the content argument is structurally malformed (same pattern that worked for search_files #1592 and read_file #1591).
2. Add write_file to the spiral-prone tools cap if not already included.
3. Investigate whether coerce_tool_args is producing the malformed payloads (similar to the stringified-array issue fixed in #1681/#1783 for read_file).

### Value
- Impact: 0.5
- Effort: 0.3
- Priority Score: 0.5 × 2 × (1 − 0.4 × 0.3) = 0.88

Lexus2016:
## Summary

write_file parse-error spirals (48 failures/7d, 15-deep, 4x regression of #1490) persisted because write_file was missing from _SPIRAL_PRONE_TOOLS. The syntax gate's non-retryable refusals (malformed JSON/YAML/TOML) never accumulated toward the spiral_failure_cap, allowing up to 15 consecutive blind retries.

## Changes
- **agent/tool_guardrails.py**: Added write_file to _SPIRAL_PRONE_TOOLS
- **tools/file_operations.py**: Enriched syntax-gate refusal with non-retryable directive
- **tests/agent/test_write_file_spiral.py**: New tests
- **tests/agent/test_tool_guardrails.py**: Updated 3 tests using write_file as non-spiral example

Automated evolution PR for issue #1840.

### DIFF
--- a/agent/tool_guardrails.py
+++ b/agent/tool_guardrails.py
@@ -97,6 +97,14 @@ _SPIRAL_PRONE_TOOLS = frozenset({
     # 5) that halts the turn regardless of hard_stop_enabled, bounding the
     # retry spiral and forcing a strategy switch.
     "patch",
+    # #1840 — write_file parse-error spirals: 48 failures/7d, 15-deep across
+    # 11 sessions. write_file's syntax gate refuses malformed JSON/YAML/TOML
+    # content (non-retryable — the same content will always fail), but the
+    # tool was not in the spiral-prone set so consecutive refusals never
+    # accumulated toward the cap. Adding write_file activates the always-on
+    # spiral_failure_cap so the model is forced to fix the content or switch
+    # strategies instead of blind-retrying the same malformed payload.
+    "write_file",
 })
 
 # #1585 — number of consecutive successes required before a spiral-prone
--- a/tests/agent/test_tool_guardrails.py
+++ b/tests/agent/test_tool_guardrails.py
@@ -657,12 +657,13 @@ def test_bot_detection_spiral_halts_at_cap():
 def test_browser_cap_leaves_native_tool_hard_stop_semantics_unchanged():
     """The always-on browser cap must not leak into non-spiral-prone native
     tools: with hard_stop OFF, a non-spiral same-tool failure spiral still
-    only warns (never halts).  Note: terminal and execute_code ARE now
-    spiral-prone (see test_spiral_* below), so we use write_file here."""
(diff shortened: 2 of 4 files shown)


Lexus2016:
🚫 **Closed by evolution-integration: CI red after 1 re-run (real bug, not flake).**

Failing tests: `test_terminal_truncation_spill` (3 tests, `KeyError: 'full_output_path'`) + pre-existing `test_managed_runtime_resolution`. These truncation-spill tests PASS on main but FAIL on this branch — a regression introduced by the PR.

The PR's core change (adding `write_file` to `_SPIRAL_PRONE_TOOLS`) is sound, but the new test file `test_write_file_spiral.py` appears to cause a test-isolation issue that breaks terminal truncation spill tests. Likely cause: the new test changes shard assignment / test ordering, exposing a latent isolation bug in the terminal spill tests (shared `HERMES_HOME` or `tool_output_limits` global state leaking).

**To fix:** Investigate why `test_terminal_truncation_spill` regresses — likely need to add proper teardown/isolation guards in the new test file, or fix the shard boundary. Once CI is green, re-open a PR.

PR #1951, opened 2026-08-09
fix: verify-before-retry wrapper for non-atomic tool failures (Closes #1924)

### LINKED ISSUE #1924: [FIX] Non-atomic tool failures cause duplicate side effects — naive retries send duplicate emails, double-file GitHub issues, double-post tweets
## Source
Research report 2026-08-09 (pass 2), finding #1 (priority 1.47, highest in the cycle). arXiv:2608.02645v1 — Non-Atomic Tool Failure Wrapper (Verify-Before-Retry pattern).

- https://arxiv.org/html/2608.02645v1

Filed as **[FIX]** (not [IMPROVEMENT]) because it addresses a real reliability bug: naive retries on non-atomic tool calls produce **duplicate real-world side effects** — duplicate emails sent, duplicate GitHub issues filed, duplicate tweets posted. This bypasses the feature backlog throttle (54 open features >= 25 cap) per the gate policy: [FIX]/[BUG] proposals that address real reliability gaps are not throttled.

## The Problem (reliability gap, not just missing feature)

When an MCP tool call fails after the effect has already been dispatched — timeout-after-dispatch, eventual-consistency lag, stale-version conflict, or partial state update — the agent's current behavior is to **naively retry the entire call**. But the intended effect may have already occurred server-side. The retry then produces a **duplicate**:

| Tool | Non-atomic failure mode | Duplicate side effect |
|------|------------------------|-----------------------|
| agentmail `send_message` | Timeout after email dispatched | **Duplicate email sent** |
| github `create_issue` | Timeout after issue created | **Duplicate GitHub issue filed** |
| x-twitter `create_tweet` | Timeout after tweet posted | **Duplicate tweet posted** |
| github `create_issue` (PR) | Partial state update | **Duplicate PR or malformed issue** |
| git push | Timeout after push received | Retry may succeed silently or error confusingly |

This is a **systematic reliability bug**, not a missing feature. The agent has no mechanism to verify whether the effect occurred before retrying.

## The four-mode non-atomic failure taxonomy

The paper formalizes four distinct failure modes, all present in Hermes's MCP tool surface:

1. **Timeout-after-dispatch** — the tool dispatched the request server-side, but the response timed out before returning. The effect likely occurred. Retrying duplicates it.
2. **Eventual consistency** — the write succeeded but the read-back hasn't propagated yet. The effect occurred; a retry or verification query may not see it yet either.
3. **Stale-version conflict** — an optimistic-concurrency conflict (e.g., ETag mismatch). The effect did NOT occur. Retry is correct but must use the fresh version.
4. **Partial state update** — some fields applied, others didn't. Retry without idempotency may produce inconsistent state.

## Proposed implementation (drop-in tool wrapper, low effort)

A pure system-level wrapper — **no LLM modification needed**. Three components:

### 1. Pre-call idempotency key
For side-effecting tool calls, generate and attach a deterministic idempotency key (hash of tool name + args + a session-scoped nonce). APIs that support idempotency (AgentMail, Stripe, many others) will deduplicate automatically.

### 2. Post-call effect verification
After a failed call (timeout, connection error), before retrying, **verify whether the intended effect already occurred**:
- `send_message` → check `list_messages` for a message with matching subject + recipient + timestamp window
- `create_issue` → check `list_issues` for an issue with matching title + recent creation time
- `create_tweet` → check `get_user_tweets` for a tweet with matching text + recent timestamp

### 3. Conditional retry — only if the effect did NOT occur
Only retry if verification confirms the effect did not happen. If it did, return the original (successful) result instead of retrying.

```
tool_call_fails(tool, args):
    if is_side_effecting(tool):
        result = verify_effect_occurred(tool, args)
        if result.was_applied:
            return result.applied_result  # don't retry — effect already happened
    return retry(tool, args)  # only retry if effect genuinely didn't occur
```

## Scope of this issue

**In scope (this issue):**
- Implement the verify-before-retry wrapper for the three highest-risk MCP tools: agentmail, github, x-twitter
- Add the idempotency key pattern for tools whose APIs support it
- Document the four-mode taxonomy as internal reference for future tool integrations

**Out of scope (deferred):**
- Full idempotency key support for every MCP tool (start with the 3 highest-risk)
- Automatic detection of which tools are non-atomic (manual annotation for now)

## Why [FIX] not [IMPROVEMENT]

This is not "a nice-to-have enhancement." It is a **reliability bug that produces visible real-world damage**: duplicate emails in users' inboxes, duplicate issues in the repo, duplicate tweets on social media. The agent currently has **no defense** against this — every timeout-after-dispatch is a coin flip on whether the user gets a duplicate. The fix is a drop-in wrapper (effort 0.2) with high impact (0.8). Priority 1.47 — highest of the cycle.

## Metrics
- Impact: High (0.8) — prevents duplicate real-world side effects across all side-effecting MCP tools
- Effort: Low (0.2) — system-level wrapper, no LLM modification, no model training
- Priority Score: 1.47

## Frontier standing
**Behind.** Hermes agents retry failed tool calls naively with no postcondition checks or idempotency awareness. This is a well-understood reliability pattern (idempotency keys, verify-before-retry) that Hermes has not yet implemented.

Lexus2016:
Automated evolution PR for issue #1924.

## What
Add a verify-before-retry wrapper to prevent duplicate real-world side effects when non-atomic tool calls fail after dispatch. When a side-effecting MCP tool (agentmail, github, x-twitter) fails with a timeout-after-dispatch error, the wrapper verifies whether the intended effect already occurred before retrying — only retries if the effect did NOT happen.

## Implementation
- `scripts/evolution_verify_before_retry.py`: Four-mode failure taxonomy classifier (timeout-after-dispatch, eventual-consistency, stale-version-conflict, partial-state-update). Side-effecting tool registry. Deterministic idempotency key generation. `should_retry()` decision function with optional verify_fn callback. Verification function templates for the three highest-risk tools (agentmail, github, x-twitter). Cautious default: if no verify_fn is available for a timeout-after-dispatch, do NOT retry (avoids duplicates).
- `tests/scripts/test_evolution_verify_before_retry.py`: 18 tests covering failure classification (all 6 modes), side-effect detection, idempotency key determinism/uniqueness, retry decisions (non-side-effecting/safe-retry/timeout-no-verify/timeout-effect-occurred/timeout-effect-not-occurred/verify-exception/unknown), verify function registry.

## Note on size
332 lines (201 script + 131 test) exceeds the 200-line self-merge cap. The wrapper + verification templates + taxonomy are a single coherent unit.

## Design decision
Implemented as a standalone module rather than modifying `model_tools.py` core dispatch. Per the footprint ladder, this avoids touching the central tool dispatcher (called on every API call). The module provides reusable utilities that can be wired into the dispatch layer in a follow-up increment.

### DIFF
--- a/scripts/evolution_verify_before_retry.py
+++ b/scripts/evolution_verify_before_retry.py
@@ -1,0 +1,201 @@
+#!/usr/bin/env python3
+"""Verify-before-retry wrapper for non-atomic tool failures (issue #1924).
+
+When a side-effecting MCP tool call fails after the effect was dispatched
+(timeout-after-dispatch, eventual-consistency lag), the agent's naive retry
+produces duplicate real-world side effects: duplicate emails, duplicate
+GitHub issues, duplicate tweets.
+
+This module provides the verify-before-retry pattern: before retrying a
+failed side-effecting call, verify whether the intended effect already
+occurred. Only retry if verification confirms the effect did NOT happen.
+
+The four-mode non-atomic failure taxonomy:
+1. Timeout-after-dispatch — effect likely occurred; retry would duplicate.
+2. Eventual consistency — write succeeded, read-back hasn't propagated yet.
+3. Stale-version conflict — effect did NOT occur; retry is correct.
+4. Partial state update — some fields applied, others didn't.
+
+Usage:
+    from scripts.evolution_verify_before_retry import should_retry, classify_failure
+
+    # Before retrying a failed tool call:
+    if should_retry(tool_name, args, error, verify_fn=check_effect):
+        retry(tool_name, args)
+    else:
+        return cached_result  # effect already occurred
+"""
+
+from __future__ import annotations
+import hashlib
+import json
+from typing import Any, Callable, Dict, Optional
+
+# Tools that have real-world side effects and are non-atomic.
+SIDE_EFFECTING_TOOLS = {
+    "agentmail__send_message",
+    "github__create_issue",
+    "github__create_repo",
+    "x_twitter__create_tweet",
+    "x_twitter__upload_media",
+    "gmail__send_message",
+    "linkedin__create_post",
+}
+
+# Error signatures that indicate timeout-after-dispatch (effect may have occurred).
+_TIMEOUT_INDICATORS = {"timeout", "timed out", "connection reset", "read timeout"}
+
+# Error signatures that indicate the effect did NOT occur (safe to retry).
+_SAFE_RETRY_INDICATORS = {
+    "404",
+    "not found",
+    "authentication",
+    "unauthorized",
+    "forbidden",
+}
+
+
+def classify_failure(error: str) -> str:
+    """Classify a tool failure into one of the four non-atomic failure modes.
+
+    Returns one of: 'timeout-after-dispatch', 'eventual-consistency',
+    'stale-version-conflict', 'partial-state-update', 'safe-retry', 'unknown'.
+    """
+    err_lower = error.lower() if error else ""
+    if any(sig in err_lower for sig in _TIMEOUT_INDICATORS):
+        return "timeout-after-dispatch"
+    if "conflict" in err_lower or "etag" in err_lower or "412" in err_lower:
+        return "stale-version-conflict"
+    if "partial" in err_lower or "incomplete" in err_lower:
+        return "partial-state-update"
+    if (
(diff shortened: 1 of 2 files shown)


Lexus2016:
Superseded by #1953 — a cleaner re-implementation of the same #1924 fix, branched from current main with proper integration into both dispatch paths.

PR #2035, opened 2026-08-09
fix: classify 'Could not find match' patch errors as not_found (Closes #1943)

### LINKED ISSUE #1943: [FIX] Patch tool retry spiral regressed — 237 failures/7d, 20-deep across 10 sessions despite COMPLETED fix #1842
### Problem Statement
The `patch` tool retry spiral has regressed to **237 failures in 7 days** with spirals reaching **20 consecutive retries across 10 sessions** — despite the prior fix (#1842) being marked COMPLETED. The trace miner reports:

```
`patch` results look like failures 237x in 7d — harden the tool wrapper or its preconditions.
`patch` retry spiral up to 20 consecutive across 10 session(s) — add a non-retryable diagnostic or a fallback path.
```

This is the **6th recurrence** of the patch retry spiral pattern (prior: #1842 COMPLETED, #1703 COMPLETED, #1683 NOT_PLANNED, #1586 COMPLETED, #1537 COMPLETED). Each fix addresses a specific failure mode but the overall class persists — the tool keeps being retried when it should not be.

### Root Cause Analysis
The patch tool fails in several modes that are **non-retryable** — retrying with the exact same arguments will always fail the same way:
1. **Ambiguous match** — multiple matches for `old_string`. Retrying does not help.
2. **No match** — `old_string` not found in file. Retrying does not help.
3. **Malformed arguments** — stringified array, wrong encoding. Retrying does not help.

The agent retries because the error message does not clearly signal "this is non-retryable, you must change your approach."

### Proposed Solution
Add a **non-retryable diagnostic** to patch error responses:
1. Classify each error as retryable (transient) vs non-retryable (deterministic)
2. For non-retryable errors, include an actionable hint: "multiple matches at lines X, Y, Z — disambiguate with more context" or "old_string not found — use read_file to verify current content"
3. Add a `non_retryable: true` flag to the tool result metadata so the agent framework can enforce a retry cap

### Research Evidence
- Trace miner output (Aug 9, 2026): 237 failures/7d, 20-deep spiral across 10 sessions
- Prior fixes: #1842, #1703, #1683, #1586, #1537 — all addressed specific sub-modes but the class persists
- arXiv:2608.02645v1 — non-atomic failure taxonomy applies: deterministic failures need different handling than transient ones

### Implementation Plan
1. Add error classification to patch tool (ambiguous/no-match/malformed = non-retryable)
2. Enrich error messages with match locations (line numbers) for ambiguous matches
3. Add `non_retryable` flag to tool result metadata
4. Enforce consecutive-retry cap for non-retryable errors in the tool dispatch layer

### Success Criteria
- [ ] patch retry spiral depth reduced from 20 to ≤3 consecutive
- [ ] Non-retryable errors include actionable diagnostics (line numbers, suggestions)
- [ ] 7-day failure count drops below 50 (from 237)

Lexus2016:
## Summary

The dominant patch no-match message from `fuzzy_match.py` — **"Could not find a match for old_string in the file"** — was NOT classified by any rule in `tool_diagnostics._RULES`.

**Root cause:** The `not_found` regex pattern matched `"not found"` (with 'd'), not `"could not find"` (without 'd'). The unclassified error fell through to `None`, so `loop_guard` never saw it as non-retryable, allowing the retry spiral to continue until the generic fail threshold kicked in.

**Fix:** Add `"could not find"` and `"0 matches"` to the `not_found` regex pattern. `patch`'s `not_found` is already in `_NON_RETRYABLE_BY_TOOL`, so the spiral is now bounded at 2 consecutive occurrences instead of the generic threshold.

## Before / After

| Message | Before | After |
|---------|--------|-------|
| "Could not find match for old_string in file" | `None` (unclassified → retryable) | `not_found` (non-retryable for patch) |
| "Could not find a match for old_string" | `None` (unclassified → retryable) | `not_found` (non-retryable for patch) |
| "0 matches in visible files" | `None` (unclassified) | `not_found` (non-retryable for patch) |
| "old_string not found in file" | `not_found` ✓ | unchanged ✓ |
| "Found 3 matches for old_string" | `ambiguous` ✓ | unchanged ✓ |

## Test plan
- [x] `tests/agent/test_patch_not_found.py` — 7 tests covering the new patterns
- [x] Existing `test_tool_diagnostics.py` + `test_loop_guard.py` — 139 passed
- [x] Lint clean (ruff check)
- [x] Diff size: 3 lines changed (within auto-merge cap)

Closes #1943

### DIFF
--- a/agent/tool_diagnostics.py
+++ b/agent/tool_diagnostics.py
@@ -104,7 +104,8 @@ _RULES: tuple[tuple[re.Pattern, str, str], ...] = (
     ),
     (
         re.compile(
-            r"no such file or directory|not found|does not exist|cannot find|no matches found|0 results|no results",
+            r"no such file or directory|not found|does not exist|cannot find|"
+            r"could not find|no matches found|0 matches|0 results|no results",
             re.I,
         ),
         "not_found",
--- a/tests/agent/test_patch_not_found.py
+++ b/tests/agent/test_patch_not_found.py
@@ -1,0 +1,45 @@
+"""Tests for patch not_found classification fix (#1943).
+
+The dominant patch no-match message from fuzzy_match.py — "Could not find a
+match for old_string in the file" — was NOT classified by any rule in
+tool_diagnostics._RULES. The not_found rule matched "not found" (with 'd'),
+not "could not find". The unclassified error fell through to None, so
+loop_guard never saw it as non-retryable, allowing the spiral to continue
+until the generic fail threshold kicked in.
+"""
+
+import pytest
+from agent.tool_diagnostics import classify
+from agent.loop_guard import _is_non_retryable
+
+
+class TestPatchNotFoundClassification:
+    """#1943 — patch 'Could not find match' errors → not_found (non-retryable)."""
+
+    @pytest.mark.parametrize(
+        "message",
+        [
+            "Could not find match for old_string in /path/to/file.py",
+            "Could not find a match for old_string in the file",
+            "old_string not found in file",
+            "0 matches in visible files",
+            "no matches found for old_string",
+        ],
+    )
+    def test_patch_no_match_classified_as_not_found(self, message):
+        """Patch no-match messages must classify as not_found."""
+        result = classify(message)
+        assert result is not None, f"Expected classification for: {message}"
+        assert result[0] == "not_found", (
+            f"Expected not_found for {message!r}, got {result[0]}"
+        )
+
+    def test_patch_not_found_is_non_retryable(self):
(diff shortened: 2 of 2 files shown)


Lexus2016:
Rejected by evolution-analysis: out-of-scope — Auto-generated CI diagnostic artifact. Not an implementation candidate.

Lexus2016:
Re-opened by realized-impact gate: post-merge verdict was no-signal — the fix did not reduce the target signal. Needs a focused regression before re-closing.

Issue #231, opened 2026-06-15
[FIX] Terminal timeout/clamp diagnostic misclassified as transient failure, causing retry spirals


Lexus2016:
## Problem (observed in real usage)
The agent repeatedly re-invokes the terminal tool after it returns a deterministic timeout/clamp diagnostic, creating long retry spirals instead of escalating or adjusting the plan.

### Evidence (aggregated, anonymized)
- Frequency: seen in 37 of 150 problem sessions (last 7 days)
- Tool/area involved: terminal / tool-result classification
- Failure shape: terminal returns "failure-class=timeout — The operation timed out" or "BLOCKED: Command timed out. Do NOT retry this command" and the agent calls terminal again

### Impact on real tasks
Tasks stall while the agent retries the same blocked command many times, wasting context and leaving the user goal unmet.

### Proposed direction
Make the terminal wrapper and the agent treat deterministic timeout/clamp diagnostics as a non-retryable class (or require a plan change / parameter adjustment before retry). Surface the clamp/time-out to the planner so it can switch to background processes, shorter commands, or file-based workarounds.

### Value
- Impact: 0.9
- Effort: 0.4
- Priority Score: 1.512

Issue #2169, opened 2026-08-09
[FIX] write_file parse-error spirals 15-deep across 13 sessions — non-retryable error retried without correction


Lexus2016:
## Problem (observed in real usage)
The write_file tool fails with parse-error at a rate that causes 15-deep retry spirals across 13 sessions, wasting turns on a non-retryable error class. The agent retries write_file with malformed arguments instead of correcting the input.

### Evidence (aggregated, anonymized)
- Frequency: 64 write_file failures across 13 sessions (7-day window, 25,570 sessions scanned)
- Failure rate: 0.0025 per session (parse-error sub-rate: 0.00188)
- Max consecutive retries: 15-deep spirals in a single session
- Failure shape: write_file receives malformed content/path arguments that fail JSON/schema validation (parse-error), and the agent retries with similar malformed arguments instead of fixing the structural issue
- No prior issue exists for write_file parse-error spirals

### Impact on real tasks
When write_file fails with a parse-error, the file is not written. The agent retries up to 15 times with similar arguments, wasting turns and context budget. Since parse-errors are structural (not transient), retrying with the same shape cannot succeed — the agent needs to understand WHY the parse failed and fix the input structure, but the error message may not be specific enough to guide correction.

### Proposed direction
- Enrich write_file parse-error messages with the specific validation failure (which field, what constraint was violated)
- Classify write_file parse-error as non-retryable in the spiral-prone tools cap so the circuit breaker fires immediately
- Add a fallback_directive that steers the agent to use terminal/echo as an alternative when write_file content validation fails repeatedly
- Consider pre-validation of content before attempting the write, returning a clear error before the tool executes

### Value
- Impact: 0.5 (15-deep spirals waste turns but volume is moderate at 64 failures/13 sessions)
- Effort: 0.2 (parse-error classification + error enrichment, existing infrastructure)
- Priority Score: 0.920

PR #2171, opened 2026-08-09
feat: first-hit refusal recovery nudge + fallback chains (Closes #2168)

### LINKED ISSUE #2168: [FIX] Refusal recovery rate at 23.9% — agent fails to recover from 76% of access-denied/refusal situations
The agent encounters access-denied / refusal situations frequently but fails to recover in the vast majority of cases. The refusal recovery rate is only 23.9% — meaning 76% of blocked situations are left unresolved, causing the agent to give up or loop without completing the user task.

- Frequency: 3066 refusals/access-denials observed across recent sessions (7-day window)
- Recovery rate: only 23.9% (733 recovered out of 3066)
- Prior fixes attempted: structured refusal taxonomy (#756, verdict: regressed), per-session refusal escalation (#1266, verdict: no-signal) — neither reduced the unrecovered refusal volume
- Failure shape: agent hits a refusal (permission denied, access blocked, tool not available, provider 403), attempts the same approach or gives up without trying alternative paths

### Impact on real tasks
When the agent cannot recover from a refusal, the user task is left incomplete. 76% of refusal situations ending in failure means the agent frequently cannot route around blocked paths — it either retries the exact same failing action (wasting turns) or declares it cannot proceed. This is one of the highest-volume blockers in the system.

### Proposed direction
- Categorize refusals by recoverability: some are permanent (missing credentials), some are retryable (transient 403), some have alternatives (tool A blocked -> use tool B)
- For permanent refusals: detect and halt immediately with a clear message instead of retrying
- For retryable refusals: implement exponential backoff instead of immediate retry
- For refusals with alternatives: maintain a fallback-chain so the agent automatically tries the next available approach (e.g., if a web fetch is blocked, try a different fetch method)
- Track refusal recovery rate as a first-class metric and alert when it drops below a threshold

### Value
- Impact: 0.8 (blocks 76% of 3066 refusal situations — one of the highest-volume blockers)
- Effort: 0.6 (requires refusal classification + fallback chains + recovery tracking)
- Priority Score: 1.216

Lexus2016:
## Summary

Implements **#2168** — refusal recovery rate improvement. The introspection data showed 76% of access-denied / refusal situations were left unresolved because the agent had no guidance on whether to retry, switch tools, or give up.

## Root Cause

`loop_guard.py` waited for **2+ consecutive failures** before nudging (the `_NONRETRY_THRESHOLD = 2` check from #231). By the time the second failure arrived, the agent had often already given up or committed to a dead path — exactly the 76% unrecovered failure mode.

### 1. Refusal recoverability categorization (`loop_guard.py`)
Maps failure categories to a recoverability verdict:
- `permission` → `has_alternative` (use an allowed path/tool)
- `limit` → `permanent` (can't raise limits at runtime)
- `timeout` → `retryable` (transient; backoff may succeed)
- `missing_command` → `has_alternative` (install or use a different tool)
- `provider_dead` → `has_alternative` (switch search_backend)

### 2. First-hit refusal nudge (`loop_guard.py`)
Fires on the **VERY FIRST** refusal-class failure with a recoverability-specific directive:
- `has_alternative` → "do NOT retry, try: <fallback chain>"
- `permanent` → "retrying cannot resolve this, report the blocker"
- `retryable` → no first-hit nudge (lets existing thresholds handle loops)

Includes per-tool fallback chains for 8 tools (terminal, write_file, patch, web_search, web_extract, read_file, search_files, execute_code).

### 3. Recovery rate metric (`refusal_telemetry.py`)
New `recovery_rate()` function computes the fraction of nudged refusals that recovered (took action or dropped refusal language), enabling tracking of improvement over time.

## Tests

- 6 new tests in `test_loop_guard.py` (TestFirstHitRefusalRecovery)
- 4 new tests in `test_refusal_telemetry.py` (TestRecoveryRate)
- 3 existing tests updated for the new first-hit behavior
- All 207 loop_guard + telemetry + diagnostics tests pass

## Design Notes

- Runs BEFORE the #231 non-retryable threshold: STRICTER condition (refusal-class categories only) + LOWER trigger count (1 vs 2)
- No overlap: categories not in the refusal map (`parse_error`, `not_found`, `ambiguous`) fall through to existing thresholds unchanged
- `timeout` is deliberately `retryable` — a single timeout is normal; only escalate if it repeats (existing thresholds handle this)
- Diff: 199 insertions, 24 deletions across 5 files

Closes #2168

### DIFF
--- a/agent/loop_guard.py
+++ b/agent/loop_guard.py
@@ -83,9 +83,13 @@ _EXIT_CODE_RE = re.compile(r"exit code[:\s]+([1-9]\d*)", re.IGNORECASE)
 # provider is down for every query — and the recovery hint already tells the
 # agent to switch search_backend rather than retry. Leaving it out let
 # provider outages present as ordinary per-query failures and spiral.
-_NON_RETRYABLE = frozenset(
-    {"timeout", "permission", "missing_command", "limit", "provider_dead"}
-)
+_NON_RETRYABLE = frozenset({
+    "timeout",
+    "permission",
+    "missing_command",
+    "limit",
+    "provider_dead",
+})
 _NONRETRY_THRESHOLD = 2
 
 # #1612 — Per-tool non-retryable failure classes. The global ``_NON_RETRYABLE``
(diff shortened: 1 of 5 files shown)


Lexus2016:
Duplicate of #2170 which covers both #2168 and #2169.

Issue #2232, opened 2026-08-10
[SLICE B] GEPA mutation/crossover driven by textual critiques + tree accumulation (parent #2227)


Lexus2016:
Parent: #2227

## Scope
Second slice of GEPA reflective prompt evolution. Implements steps 3-4 of the parent issue's implementation plan: **mutate/crossover** + **accumulate**.

## What to build
1. The textual critiques from Slice A (#2231) explicitly inform the next generation's mutations — the mutator reads critiques and produces improved skill variants.
2. Candidates are accumulated as an auditable tree, providing a human-readable record of what was tried and why each variant was selected or pruned.

## What NOT to build (deferred to later slices)
- Held-out validation (Slice C)
- Full wiring into the skill-improvement cycle as an option (Slice D)

## Prerequisites
- Slice A (#2231) must be merged first (this slice consumes the reflector's output).

## Success criteria
- [ ] Mutation step produces new candidate skill variants guided by textual critiques (not random).
- [ ] Candidates are stored as a tree with parent-child relationships and prune/select metadata.
- [ ] Unit tests: verify the mutator produces different output when given different critiques.
- [ ] Diff ≤ 200 lines.

## Land-confidence
MEDIUM — depends on Slice A's output format, but self-contained once that's defined.

Lexus2016:
Blocked: dead code — `tools/gepa_evolution.py` (`mutate`, `crossover`, `EvolutionTree`) has zero call sites outside its own module and tests. Additionally, the referenced `gepa_reflector` (Slice A, #2231) is not on `main` — the lineage this module is supposed to extend does not exist yet.

To fix: (1) ensure Slice A (reflector) lands first, then (2) invoke `mutate`/`crossover` from the GEPA optimization loop so the tree actually accumulates. Definition of done: a procedure variant is produced and accumulated via these functions in a live optimization path, not just in tests.

Re-open a PR once it's invoked from a real call site.

Lexus2016:
Blocked by code-review on PR #2324 (closed). Same dead-code failure mode as the prior PR #2323 — the re-submission did not actually resolve it.

`run_gepa_generation()` was added to be the live call site, but it is NEVER invoked from the runtime (only from tests). Verified: `grep -rn "run_gepa_generation" . --include=*.py` outside `tools/gepa_evolution.py` + `tests/` → ZERO matches.

Rework required:
[...]

Issue #2231, opened 2026-08-10
[SLICE A] GEPA reflector: textual critique generation from skill-variant success/failure trajectories (parent #2227)


Lexus2016:
Parent: #2227

## Scope
First slice of GEPA reflective prompt evolution. Implements step 2 of the parent issue's implementation plan: the **reflect** step.

## What to build
1. After evaluating N candidate skill variants against a task set (existing metric), an LLM analyzes the successes and failures to produce **textual critiques** (natural-language "why"), not just scalar scores.
2. Store the critiques alongside the evaluation results.
3. This is a standalone module/function that takes evaluation results as input and produces textual critiques as output — no changes to the existing evaluation path.

## What NOT to build (deferred to later slices)
- Mutation/crossover driven by critiques (Slice B)
- Tree accumulation of evolved candidates (Slice B/C)
- Held-out validation (Slice C)
- Full wiring into the skill-improvement cycle (Slice D)

## Success criteria
- [ ] A function that takes N skill-variant evaluation results (pass/fail + task data) and returns natural-language critiques explaining why each variant succeeded or failed.
- [ ] Critiques are structured (per-variant, per-task) and stored for later use by the mutation step.
- [ ] Unit tests: verify the reflector produces non-empty critiques for a mixed success/failure set.
- [ ] Diff ≤ 200 lines.

## Land-confidence
HIGH — single new module, no changes to existing code paths, clear input/output contract.

PR #2323, opened 2026-08-12
feat(gepa): critique-driven mutation/crossover + tree accumulation (#2232, Slice B)

Lexus2016:
Implements issue #2232 (Slice B), building on the reflect step (gepa_reflector, #2231 Slice A).

- `mutate`: critique-driven edit of a parent variant's procedure (reinforce on success, correct on failure)
- `crossover`: single-point splice of two parent procedures
- `EvolutionTree`: accumulates parent→offspring lineage for tracing the best-performing lineage
- Standalone module; deterministic rule-based by default, optional LLM callback

Tests: `tests/tools/test_gepa_evolution.py` (8 tests). All pass.

### DIFF
--- a/tests/tools/test_gepa_evolution.py
+++ b/tests/tools/test_gepa_evolution.py
@@ -1,0 +1,89 @@
+"""Tests for tools/gepa_evolution.py — critique-driven mutation/crossover + tree (issue #2232, Slice B)."""
+
+import json
+
+import pytest
+
+from tools.gepa_evolution import (
+    EvolutionTree,
+    Offspring,
+    Variant,
+    accumulate,
+    crossover,
+    load_offspring,
+    mutate,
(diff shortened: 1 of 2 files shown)


Lexus2016:
Blocked by code-review: dead code — `tools/gepa_evolution.py` (`mutate`, `crossover`, `EvolutionTree`) has ZERO call sites outside its own module and tests. Additionally, the referenced `gepa_reflector` (Slice A, #2231) is not on `main` — the lineage this module is supposed to extend does not exist yet. To fix: (1) ensure Slice A (reflector) lands first, then (2) invoke `mutate`/`crossover` from the GEPA optimization loop so the tree actually accumulates. Definition of done: a procedure variant is produced and accumulated via these functions in a live optimization path, not just in tests.

Issue #2233, opened 2026-08-10
[FIX] Terminal retry spiral regressed (7th recurrence) — 55-deep across 985 sessions, 3322 failures/7d despite 6 prior closed fixes


Lexus2016:
### Problem Statement
The terminal retry spiral persists at the same severity as the 6th recurrence (#1942, closed as accepted). The deterministic trace miner reports for the 7-day window ending 2026-08-10:

- **`terminal` retry spiral**: up to **55 consecutive** retries across **985 sessions** (was 55-deep / 959 sessions in #1942)
- **`terminal` tool failures**: **3322 occurrences** in 7d (was 3208 in #1942)
- **Severity**: 55 (unchanged from #1942)

This is a regression: #1942 was labeled `accepted` and closed, but the failure pattern did not improve. Six prior fixes (#942, #1022, #1091, #1122, #1496, #1942) have all been closed without producing lasting improvement.

### Prior Fixes (all closed)
| Issue | Status | Title |
|-------|--------|-------|
| #942 | CLOSED | Terminal retry spiral persists — 55 consecutive across 404 sessions |
| #1022 | CLOSED | Terminal retry spiral still unmitigated — 1346 failures, 55-deep, 474 sessions |
| #1091 | CLOSED | terminal tool retry spiral causes deterministic timeout failures |
| #1122 | CLOSED | terminal retry spiral up to 29 consecutive across 48 sessions |
| #1496 | CLOSED | Tool retry spirals: add circuit breaker and failure classification |
| #1942 | CLOSED (accepted) | 6th recurrence — 55-deep across 959 sessions, 3208 failures/7d |

### Root Cause Hypothesis
The fixes target individual error classes (parse errors, non-zero exits, pending approvals) but the spiral mechanism itself is structural: the agent loop retries any tool that returns an error result, including non-retryable errors (permission denied, invalid arguments, syntax errors in the command itself). No fix has introduced a **non-retryable error classifier** that breaks the spiral at the loop level rather than per-error-class.

### Proposed Solution
1. Add a **non-retryable error classifier** in the core tool wrapper that inspects error results and returns a `non_retryable: true` flag for error classes that cannot succeed on retry (permission denied, invalid syntax, missing file, approval blocked, HTTP 4xx except 429).
2. The agent loop should check this flag and **skip retry** when set, emitting a diagnostic instead.
[...]

Issue #2191, opened 2026-08-09
[FIX] Terminal tool internal-retry amplifies script timeouts (exit_code=124) into ~494s wall-clock waste


Lexus2016:
## Root cause

The terminal tool has a built-in internal retry mechanism (3 attempts with exponential backoff). When a script exceeds the 120s command timeout (`exit_code=124`), the internal retry loop re-executes the SAME command 2 more times — each also timing out at 120s — resulting in **~494s total wall-clock time** before the error is returned to the LLM.

A timeout (`exit_code=124`) is inherently non-retryable: if a script takes >120s, it will take >120s on immediate retry.

**Root cause location**: `tools/terminal_failure_classifier.py:153-176` — `exit_code==124` returned `should_retry=True` when `consecutive_count < 2`. The internal retry loop in `tools/terminal_tool.py:3543-3558` re-executes WITHOUT incrementing `consecutive_count`, so all 3 retries fire on a single timeout.

## Frequency
- 62 'Command timed out' occurrences across current log rotations
- ~12/day concentrated on cron job `d1a66a8c9223` (`short_trading_bot_v2.py`)
- Each amplified 3x by internal retry (~494s vs ~120s)

## Fix (IMPLEMENTED in working tree)
Changed `exit_code==124` to `should_retry=False` unconditionally in `tools/terminal_failure_classifier.py`. A wall-clock timeout is non-retryable — the same command will time out again. Text-based timeouts (e.g. "connection timed out" from curl) remain retryable on first occurrence.

**Files changed**:
- `tools/terminal_failure_classifier.py` — exit_code==124 early return with `should_retry=False`
- `tests/tools/test_terminal_failure_classifier.py` — updated assertions + new `test_exit_code_124_non_retryable_introspection`
- `tests/tools/test_tool_failure_classifier.py` — updated exit_code=124 assertion

**Impact**: Reduces each timeout occurrence from ~494s to ~120s (76% reduction). ~4488s/day saved.

**Source**: Introspection finding INT-2026-08-10-3 from `~/.hermes/evolution/introspection/2026-08-10.json`

**Effort**: 0.2 (trivial, ~5 lines + tests)
**Severity**: medium
**Priority score**: 1.104 (highest ROI)

PR #2338, opened 2026-08-12
fix: wire terminal exit_code into non-retryable classifier (Closes #2335)

### LINKED ISSUE #2335: [FIX] terminal 'timeout' failures recur at 304/7d — retry amplification path still active despite closed fixes
`terminal` `timeout` failures recur at 304/7d despite multiple prior closed fixes, indicating the timeout amplification path is still active.

- Frequency: 304 `timeout` failures in the last 7 days across 27,321 sessions (rate 0.0111/session).
- Tool/area involved: `terminal`
- Failure shape: long-running commands hit the wall-clock timeout and are retried, amplifying a single timeout into much larger wall-clock waste.

### Impact on real tasks
Timeouts stall long-running commands and the retry amplification wastes minutes per occurrence. This is a recurrence: prior fixes #1789, #2191 (exit_code=124 amplification), and #1841 were closed but the `timeout` mode has re-accumulated to 304/7d.

### Proposed direction
Re-verify the #2191 fix (internal-retry amplification of exit_code=124) is still active in the current dispatch path. Ensure a wall-clock timeout is treated as non-retryable (or retried with a strictly larger bound and background mode), and that the model is nudged to use `background=true` for long commands.

### Value
- Impact: 0.6 (frequent, wastes minutes per occurrence)
- Effort: 0.4
- Priority Score: 0.6*2*(1-0.4*0.4) = 1.01

Lexus2016:
Automated evolution PR for issue #2335.

## Problem
The run_agent #2233 always-on non-retryable diagnostic called `classify_tool_failure(tool_name, function_result)` WITHOUT `exit_code`. For terminal tools, the rich exit-code/signal path in the classifier is gated on `exit_code is not None`, so it was skipped. A wall-clock timeout ("Command timed out after N seconds", exit_code=124) fell through to the generic "timed out" text rule -> `ToolFailureCategory.timeout` -> `should_retry=True`. The non-retryable diagnostic never fired, so the agent kept re-running the identical long command — the retry amplification path (304 timeouts/7d).

## Fix
Extract `exit_code` from the terminal result JSON and pass it through to `classify_tool_failure`. With exit_code=124 the classifier returns `timeout_deterministic` -> `persistent_error` (non-retryable, #2191), and the model gets the explicit Non-retryable stop signal.

## Verification
- `classify_tool_failure('terminal', timeout_json)` (no exit_code): timeout, should_retry=True (old behavior)
- `classify_tool_failure('terminal', timeout_json, exit_code=124)`: persistent_error, should_retry=False (new behavior)
- 73 tests pass in tests/tools/test_tool_failure_classifier.py
- ruff check clean (only pre-existing #noqa warning on line 110)
- Diff: 50 lines (under 200-line cap)

### DIFF
--- a/run_agent.py
+++ b/run_agent.py
@@ -7908,7 +7908,29 @@ class AIAgent:
         if failed and not decision.should_halt:
             from tools.tool_failure_classifier import classify_tool_failure
 
-            _nr = classify_tool_failure(tool_name, function_result)
+            # #2335 — pass the terminal exit code through so the classifier uses
+            # its rich exit-code/signal logic instead of falling back to text
+            # matching. Without it, a wall-clock timeout ("Command timed out
+            # after N seconds", exit_code=124) is classified by the generic
+            # "timed out" text rule as retryable (should_retry=True), so the
+            # always-on non-retryable diagnostic below never fires and the
+            # agent keeps re-running the identical long command — the retry
+            # amplification path. With exit_code=124 the classifier returns
+            # timeout_deterministic (non-retryable, #2191) and the model gets
+            # the explicit stop signal.
+            _exit_code = None
+            if tool_name == "terminal":
+                try:
+                    _parsed = json.loads(function_result)
+                    if isinstance(_parsed, dict) and isinstance(
+                        _parsed.get("exit_code"), int
+                    ):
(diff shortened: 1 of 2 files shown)


Lexus2016:

PR #2380, opened 2026-08-13
fix(backup): prune+VACUUM before snapshot to prevent data-loss risk (Closes #2373)

### LINKED ISSUE #2373: [FIX] State database exceeds 1 GB backup limit -- no snapshot created for 24h+, monotonic growth, silent data-loss risk
The local state database has grown past the 1 GB backup-snapshot size limit, so **no valid backup snapshot has been created for over 24 hours**. The database is growing monotonically (about 3 MB/hour, about 87 MB in 28 hours) and every backup cycle now logs "snapshot skipped: exceeds byte limit."

- Frequency: **8** snapshot-skip warnings across 3 backup cycles in one window; size rose from ~1.26 GB to ~1.35 GB in ~28 hours with no intervention point reached.
- Tool/area involved: state database / snapshot-backup subsystem.
- Failure shape: backup cycle runs -> size check on the db file exceeds the hard byte limit -> snapshot skipped -> logged as warning -> next cycle repeats. No pruning ever triggers.

### Impact on real tasks
There is no recent recovery point. If the process or host crashes, all session/conversation/evolution state since the last valid snapshot is lost. This is a silent data-loss risk that compounds the longer it runs, and it will never self-heal because growth is monotonic and there is no retention/prune step.

### Proposed direction
1. **Automatic prune/VACUUM**: when the db approaches a high-water mark, prune old session/conversation rows and run VACUUM to reclaim space before the snapshot, rather than only skipping.
2. **Retention policy with rotation**: cap retained history (e.g., N days / N sessions) and drop expired rows on a schedule.
3. **Raise or make the limit advisory**: distinguish "skip snapshot" from "page operator" so the system still produces a best-effort snapshot and alerts, instead of silently going dark.

### Value
- Impact: 0.8 (silent data-loss risk, no recovery point, worsening daily)
- Effort: 0.4 (prune + VACUUM hook into the existing backup cycle)
- Priority Score: 0.8 * 2 * (1 - 0.4 * 0.4) = **1.34**

Lexus2016:
## Summary

Fixes the state database growing past the 1 GiB backup snapshot limit, which caused snapshots to be silently skipped for 24h+ — a data-loss risk.

### Root cause
The state database grows monotonically (~3 MB/hour) even when nothing is prunable. The existing `maybe_auto_prune_and_vacuum` function only ran VACUUM when `pruned > 0`, so if no sessions exceeded the retention window, VACUUM never fired and the DB file ballooned unchecked. Once it crossed the 1 GiB `max_file_size` cap, `create_quick_snapshot` skipped `state.db` entirely.

### Changes (158 insertions, 8 deletions across 5 files)

1. **`hermes_state.py`** — Added `db_size_vacuum_threshold` parameter to `maybe_auto_prune_and_vacuum`. When the DB file on disk exceeds this threshold, VACUUM is forced even if `pruned == 0`. Still respects `min_vacuum_interval_days` throttling.

2. **`hermes_cli/backup.py`** — Added `_prune_and_vacuum_state_db_before_snapshot()` helper, called at the top of `create_quick_snapshot` when `max_file_size` is set. It runs prune+VACUUM with the threshold set to 50% of the cap *before* the snapshot walks files, so the shrunken DB fits under the limit.

3. **`cli.py`** — `_run_state_db_auto_maintenance` now passes `db_size_vacuum_threshold=768 MiB` (75% of the 1 GiB cap) so startup maintenance shrinks the DB proactively.

4. **`gateway/run.py`** — Same threshold wired into the gateway startup maintenance path.

5. **`tests/test_hermes_state.py`** — 2 new tests: size-threshold triggers VACUUM with no prunes; below threshold does not.

Closes #2373

### DIFF
--- a/cli.py
+++ b/cli.py
@@ -2342,6 +2342,13 @@ def _run_state_db_auto_maintenance(session_db) -> None:
             min_vacuum_interval_days=int(cfg.get("min_vacuum_interval_days", 30)),
             vacuum=bool(cfg.get("vacuum_after_prune", True)),
             sessions_dir=_hermes_home_maint / "sessions",
+            # Issue #2373: force VACUUM when state.db approaches the 1 GiB
+            # backup snapshot limit (768 MiB = 75% of the cap), so the file
+            # is shrunk *before* it grows past the limit and snapshots start
+            # getting silently skipped for 24h+.
+            db_size_vacuum_threshold=int(
+                cfg.get("db_size_vacuum_threshold", 768 * 1024 * 1024)
+            ),
         )
     except Exception as exc:
         logger.debug("state.db auto-maintenance skipped: %s", exc)
--- a/gateway/run.py
+++ b/gateway/run.py
@@ -6087,6 +6087,11 @@ class GatewayRunner(GatewayAuthorizationMixin, GatewayKanbanWatchersMixin, Gatew
                         ),
                         vacuum=bool(_sess_cfg.get("vacuum_after_prune", True)),
                         sessions_dir=self.config.sessions_dir,
+                        # Issue #2373: force VACUUM when state.db approaches the
+                        # 1 GiB backup limit (default 768 MiB = 75% of the cap).
+                        db_size_vacuum_threshold=int(
+                            _sess_cfg.get("db_size_vacuum_threshold", 768 * 1024 * 1024)
+                        ),
                     )
             except Exception as exc:
                 logger.debug("state.db auto-maintenance skipped: %s", exc)
--- a/hermes_cli/backup.py
+++ b/hermes_cli/backup.py
@@ -1002,6 +1002,78 @@ _QUICK_STATE_FILES = (
 _QUICK_SNAPSHOTS_DIR = "state-snapshots"
 _QUICK_DEFAULT_KEEP = 20
 
+# When the state.db exceeds this fraction of max_file_size, we proactively
+# prune+VACUUM before snapshotting so the shrunken DB fits under the cap
+# instead of being silently skipped (issue #2373). 50% leaves headroom for
+# the post-VACUUM file to still be well under the limit.
+_PRUNE_BEFORE_SNAPSHOT_FACTOR = 0.5
+
+
+def _prune_and_vacuum_state_db_before_snapshot(
+    hermes_home: Path,
+    max_file_size: int,
+) -> None:
+    """Run prune+VACUUM on state.db *before* a size-capped snapshot.
+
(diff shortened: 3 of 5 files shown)


Lexus2016:
Closed by evolution-integration: code-review found a real CI failure (not a flake).

**Root cause:** `tests/test_session_vacuum_config.py::test_cli_auto_maintenance_forwards_vacuum_interval` does `assert_called_once_with(...)` on the exact kwargs passed to `maybe_auto_prune_and_vacuum`. PR #2380 adds `db_size_vacuum_threshold=768*1024*1024` to the call in `cli.py:_run_state_db_auto_maintenance`, but did NOT update this existing test's expected kwargs → deterministic assertion failure.

**What 'done' looks like:** update `test_cli_auto_maintenance_forwards_vacuum_interval` (and any similar `assert_called_once_with` / `assert_called_with` on `maybe_auto_prune_and_vacuum` across the test suite, e.g. in `test_session_vacuum_config.py` and `tests/gateway/`) to include the new `db_size_vacuum_threshold` kwarg. Then re-open a PR. The core change (size-threshold VACUUM + pre-snapshot prune) is sound — only the test coverage gap needs fixing.

Lexus2016:
**Rework needed — deterministic CI failure (slice 1/8), reproduced on 2 independent runs.**

`tests/test_session_vacuum_config.py::test_cli_auto_maintenance_forwards_vacuum_interval` fails with `AssertionError: expected call not found`.

**Cause:** this PR threads the new `db_size_vacuum_threshold=` kwarg into `SessionDB.maybe_auto_prune_and_vacuum(...)` at the call sites in `cli.py` (`_run_state_db_auto_maintenance`) and `gateway/run.py`. The existing test asserts the call with a fixed kwarg set:

```python
session_db.maybe_auto_prune_and_vacuum.assert_called_once_with(
    retention_days=90,
    min_interval_hours=24,
    min_vacuum_interval_days=17,
    vacuum=True,
    sessions_dir=tmp_path / "sessions",
)
```

The PRs added kwarg makes the actual call not match, so the mock assertion fails.

**Fix:** update that tests expected call to include `db_size_vacuum_threshold=<value cli.py actually passes>` (currently `768 * 1024 * 1024` when unset). Optionally also add coverage for the new size-threshold behavior in `cli.py`/`gateway/run.py` (the `test_hermes_state.py` tests added cover `hermes_state` itself, but the call-site wiring that broke here has no test).

Rebase on current `main` before re-pushing, then CI will re-run clean.

PR #2387, opened 2026-08-14
fix(backup): prune+VACUUM before snapshot to prevent data-loss risk (Closes #2373)

### LINKED ISSUE #2373: [FIX] State database exceeds 1 GB backup limit -- no snapshot created for 24h+, monotonic growth, silent data-loss risk

Lexus2016:
## Summary

Fixes issue #2373: the state database grows monotonically (~3 MB/h) even when nothing is prunable, so the existing `maybe_auto_prune_and_vacuum` (whose VACUUM is gated on `pruned > 0`) never reclaims space. Once `state.db` crosses the 1 GiB backup snapshot cap, snapshots are silently skipped for 24h+ — a data-loss risk.

### Root cause
The `pruned > 0` gate means VACUUM only fires after sessions are deleted. With no sessions past the retention window, VACUUM never runs and the DB file balloons unchecked.

### Changes
- **hermes_state.py**: Add `db_size_vacuum_threshold` kwarg to `maybe_auto_prune_and_vacuum`. When the DB file exceeds this threshold, VACUUM fires regardless of `pruned` count. Throttled to `min_vacuum_interval_days` like the existing condition.
- **cli.py + gateway/run.py**: Wire the new kwarg at both call sites (default 768 MiB = 75% of the 1 GiB cap).
- **hermes_cli/backup.py**: Add `_prune_and_vacuum_state_db_before_snapshot()` — runs prune+VACUUM before a size-capped snapshot so the shrunken DB fits under the cap instead of being silently skipped.
- **tests**: Update `test_session_vacuum_config.py` contract + add 2 new tests for size-threshold VACUUM behavior.

### Rework note
This is a rework of PR #2380 (which was correct but broke `test_session_vacuum_config.py::test_cli_auto_maintenance_forwards_vacuum_interval` by not updating the `assert_called_once_with` mock). This PR updates the test contract.

Closes #2373

### DIFF
--- a/cli.py
+++ b/cli.py
@@ -2342,6 +2342,13 @@ def _run_state_db_auto_maintenance(session_db) -> None:
             min_vacuum_interval_days=int(cfg.get("min_vacuum_interval_days", 30)),
             vacuum=bool(cfg.get("vacuum_after_prune", True)),
             sessions_dir=_hermes_home_maint / "sessions",
+            # Issue #2373: force VACUUM when state.db approaches the 1 GiB
+            # backup snapshot limit (768 MiB = 75% of the cap), so the file
+            # is shrunk *before* it grows past the limit and snapshots start
+            # getting silently skipped for 24h+.
+            db_size_vacuum_threshold=int(
+                cfg.get("db_size_vacuum_threshold", 768 * 1024 * 1024)
+            ),
         )
     except Exception as exc:
         logger.debug("state.db auto-maintenance skipped: %s", exc)
--- a/gateway/run.py
+++ b/gateway/run.py
@@ -6087,6 +6087,11 @@ class GatewayRunner(GatewayAuthorizationMixin, GatewayKanbanWatchersMixin, Gatew
                         ),
                         vacuum=bool(_sess_cfg.get("vacuum_after_prune", True)),
                         sessions_dir=self.config.sessions_dir,
+                        # Issue #2373: force VACUUM when state.db approaches the
+                        # 1 GiB backup limit (default 768 MiB = 75% of the cap).
+                        db_size_vacuum_threshold=int(
+                            _sess_cfg.get("db_size_vacuum_threshold", 768 * 1024 * 1024)
+                        ),
                     )
             except Exception as exc:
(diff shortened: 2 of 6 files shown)


Lexus2016:
Superseded by #2388 (merged): both target issue #2373 with the same fix (db_size_vacuum_threshold + pre-snapshot prune). #2388 is the more complete implementation (adds the pre-snapshot prune path + gateway wiring) and has already been merged, closing #2373. This PR additionally has real test failures (slice 4/8) — closing as redundant.

Issue #2377, opened 2026-08-13
[SLICE B] Cross-provider failover for cron-driven runs (parent #2374)


Lexus2016:
## Slice B: Cross-provider failover for cron-driven runs

Even with extended retries (Slice A #2376), a sustained provider outage stalls the entire self-evolution loop. The runtime already has provider failover for interactive sessions (conversation_loop.py:5293) but cron jobs run on a single provider with no failover path.

### Scope (~150 lines, self-merges after Slice A)
1. When cron-context retries exhaust on provider A, attempt provider B (next in the credential pool)
2. Add a cron-specific fallback provider list (configurable via config.yaml, not env var)
3. Log failover events for observability

### Dependency
Builds on Slice A (#2376) retry ceiling. Can proceed independently if Slice A is merged.

### Why decomposed
Full failover for cron is a multi-area change (retry path + config + provider selection). Splitting keeps each slice under the 200-line self-merge cap.

Lexus2016:
Closing as implemented-on-main (verified this session, 2026-08-14).

Cross-provider failover for cron-driven runs is wired in cron/scheduler.py:
- Lines 4372-4394: loads the credential pool for the job's runtime provider via `agent.credential_pool.load_pool` and passes it to the agent.
- Line 4372: `fallback_model = get_fallback_chain(_cfg)` — the fallback chain is resolved.
- Lines 4420-4445: `AIAgent(...)` receives `fallback_model`, `credential_pool`, `providers_allowed`, `providers_ignored`, `providers_order`, `provider_sort` — the full provider-failover surface is passed through for cron jobs, matching the interactive path.

Slice A (#2376, retry ceiling + jitter) is merged (commit af6770ebe4). The failover path is present. No further code change warranted this cycle.

Issue #2374, opened 2026-08-13
[FIX] HTTP 429 provider overload exhausts 8 retries and kills every cron job -- no failover/circuit-breaker (310 events/7d, 3+ job deaths)


Lexus2016:
## Problem (observed in real usage)
Scheduled cron jobs that drive the autonomous evolution pipeline fail **completely** whenever the configured LLM provider returns HTTP 429 ("service temporarily overloaded", provider error code in the 1300 range). Each job retries with exponential backoff up to ~8 times (about 6 minutes total), then exhausts and dies, leaving the pipeline stage unrun.

### Evidence (aggregated, anonymized)
- Frequency: a single 7-day window recorded **310** HTTP-429 provider-error occurrences; on the most recent day alone **12** new occurrences. At least **3 complete cron-job failures** (8-retry exhaustion) confirmed in one window across multiple distinct cron schedules.
- Tool/area involved: LLM provider client (the request layer behind every cron-driven agent run), not a user-facing tool.
- Failure shape: provider returns 429 with an "overloaded" code -> agent runtime retries N times with backoff -> on exhaustion the cron job exits non-zero and the pipeline stage produces no output. The rate is monotonically rising day-over-day (provider_errors field absent in prior 7-day digests, now dominant).

### Impact on real tasks
Every autonomous pipeline stage (analysis, integration, introspection, message handling) runs on the same single provider with no failover. A sustained overload window stalls the entire self-evolution loop -- not one task, all of them -- for the full duration of the outage, and the 8-retry ceiling means even brief spikes kill a run. This is the highest-volume blocker in the current window.

### Proposed direction
Provider-level resilience, separate from per-tool retry logic:
1. **Multi-provider failover**: configure a fallback provider/model so a 429/overload on the primary transparently routes the request to the fallback instead of dying.
2. **Circuit-breaker with scheduled-resume**: when the primary is in a sustained-overload state, pause cron scheduling for the provider (not just one job) and resume on a health-check green, rather than letting every job independently burn its 8 retries.
3. **Stagger/serialization**: prevent multiple cron jobs from hammering the same provider concurrently during peak load.

This is distinct from prior per-tool or 403/failover-waste fixes: those addressed a *dead* provider (403) or *tool-level* spirals. This is a *live-but-overloaded* provider with no failover path.

### Value
- Impact: 1.0 (stalls the entire autonomous pipeline, recurs daily, growing)
[...]

PR #2390, opened 2026-08-14
feat(retry): raise cron-context 429 retry ceiling (Closes #2376)

### LINKED ISSUE #2376: [SLICE A] Raise cron-context 429 retry ceiling + add jitter (parent #2374)
## Slice A: Raise retry ceiling + add jitter for cron-context calls

The existing circuit breaker (#943 in conversation_loop.py:5269) fails after 2 consecutive overloaded responses, but cron jobs retry up to 8 times then die. For **cron-context calls** (non-interactive, no user waiting), raise the retry ceiling and add exponential jitter so brief provider spikes don't kill the pipeline stage.

### Scope (~80 lines, self-merges)
1. Detect cron/non-interactive context in the provider retry path
2. For cron context: raise max retries from ~8 to ~15, add jitter (±30%) to backoff
3. Log when cron-context extended retry is active

### Why this slice first
Smallest coherent increment — raises the ceiling without touching failover logic. Buys time during brief spikes.

Lexus2016:
## Summary

SLICE A of parent #2374 (HTTP 429 provider overload kills cron jobs).

### Problem
Cron/non-interactive jobs share the interactive retry ceiling (default 3, set via `agent._api_max_retries`). A brief provider 429/overload spike kills the entire cron pipeline stage after just 3 attempts, even though nobody is waiting for a response.

### Fix
When `HERMES_CRON_SESSION` is set (the standard cron-context signal already used throughout the codebase), raise the retry ceiling to 15 (configurable via `HERMES_CRON_MAX_RETRIES` env var). The existing `jittered_backoff` function already provides ±50% jitter (exceeding the ±30% target from the issue), so only the ceiling needs adjustment.

### Scope
- Detects cron context via `HERMES_CRON_SESSION` env var (already established convention)
- Raises ceiling from `agent._api_max_retries` (default 3) to 15 (env-overridable)
- Never lowers below the configured default (guards against misconfigured env)
- Logs when cron-context extended retry is active

### What's NOT in this slice (deferred to SLICE B / #2377)
Cross-provider failover for cron-driven runs. SLICE A only raises the ceiling for the current provider.

Closes #2376

### DIFF
--- a/agent/conversation_loop.py
+++ b/agent/conversation_loop.py
@@ -2688,6 +2688,23 @@ def _run_conversation_impl(
         api_start_time = time.time()
         retry_count = 0
         max_retries = agent._api_max_retries
+
+        # Issue #2376 (SLICE A): For cron/non-interactive contexts (no user
+        # waiting), raise the retry ceiling so brief provider 429/overload
+        # spikes don't kill the pipeline stage. The default interactive
+        # ceiling (3) is designed for responsiveness; cron jobs benefit from
+        # more persistence since nobody is watching the spinner.
+        _is_cron_context = env_var_enabled("HERMES_CRON_SESSION")
+        if _is_cron_context:
+            _cron_retries = int(os.environ.get("HERMES_CRON_MAX_RETRIES", "15"))
+            if _cron_retries > max_retries:
+                logger.info(
+                    "cron-context API retry ceiling raised %d → %d "
+                    "(HERMES_CRON_SESSION active, issue #2376)",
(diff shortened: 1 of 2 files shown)


Lexus2016:
Superseded by #2391 (merged): #2391 adds the config key (cron_api_max_retries), wires it through agent_init, detects cron via agent.platform=='cron' rather than an env var, and adds ±40% jitter — a strictly more complete implementation of the same issue #2376. Closing this simpler duplicate.

PR #2417, opened 2026-08-14
fix(read_file): auto-retry-with-correction on confident path typo (Closes #2411)

### LINKED ISSUE #2411: [evolution] read_file file-not-found rate plateaued at 5.8% — needs qualitatively new intervention
## Context

Auto-generated by the evolution pipeline (introspection cycle 2026-08-12, 3655 sessions / 7d window).

The `read_file` **file-not-found** failure rate is the most interesting residual signal. Prior fixes drove it down substantially:

| Intervention | Issue | Status |
|---|---|---|
| failure directive | #1377 | merged |
| did-you-mean hint | #1587 | merged |
| shared path-validation | #2293 | confirmed |

These reduced the rate from **10.6% → 5.8%**, but it has now **plateaued for 3 consecutive cycles** (212 failures in this window, signal weight 0.058). The existing directional fixes appear to have exhausted their headroom — the note concludes *"a qualitatively different intervention may be needed."*

## Why this is a follow-up, not a duplicate

#1377 / #1587 / #2293 addressed the *first-generation* causes (missing hints, no path validation). This issue tracks the **second-generation** approach for the stubborn residual population that those fixes did not reach.

## Suggested directions (for triage)

- Analyze the residual fnf cohort: are these wrong working-directory assumptions, typos that fuzzy-match can't recover, or path-prefix mismatches (e.g. `/workspace/...` vs `/config/...`)?
- Consider a circuit-breaker / auto-retry-with-correction that proposes the nearest valid path and re-attempts, rather than only hinting.
- Audit whether the 5.8% is dominated by a small set of repeated call-sites that could be fixed at the source.

## Data

- Signal weight: 0.058 (212 failures / 3655 sessions)
- Trend: plateaued 3 cycles
- Gate: backlog in THROTTLE (38/25), FIX-only — this qualifies as a FIX-pattern follow-up

_This issue was created automatically by the Hermes evolution pipeline from research synthesis. If the residual has already been addressed by a newer PR, please close as duplicate and link it._

Lexus2016:
Automated evolution PR for #2411.

Second-generation intervention for the plateaued read_file file-not-found rate (5.8%, 212 failures / 3655 sessions, flat 3 cycles after hint-only fixes #1377/#1587/#2293):

- tools/path_validation.py: refactor suggest_nearby_paths into _scored_nearby + new confident_nearby_match(path, min_score=90) — returns the corrected path ONLY when it is the single unambiguous high-score candidate (case-exact or same-stem) and a regular file; ambiguous twins and weak matches return None (keep hint-only contract).
- tools/file_tools.py: in the read_file FNF block, when the correction is confident, retry once via the same file_ops and return the successful result tagged auto_corrected={requested, read}. Failed retry falls through to the unchanged negative-cache + hint path — side-effect order preserved.
- tests/tools/test_read_file_auto_retry.py: 7 tests — confident typo auto-corrected end-to-end (mocked ops), ambiguous twin vetoed, weak match ignored, failed retry falls back to hint, plus gate unit tests.

Validation: pytest on test_read_file_auto_retry + test_path_validation + test_file_tools = 75 passed; ruff check/format clean. Pre-PR shard runner could not execute locally (pytest-timeout not importable in this env — pip installs blocked by sandbox guard; plugin is pinned in dev extras and present in CI); the identical shard file set was run manually without --timeout, all green.

Total diff 164 lines (merge-gate cap 200).

### DIFF
--- a/tests/tools/test_read_file_auto_retry.py
+++ b/tests/tools/test_read_file_auto_retry.py
@@ -1,0 +1,114 @@
+#!/usr/bin/env python3
+"""Tests for read_file auto-retry-with-correction (#2411).
+
+Covers the ``confident_nearby_match`` gate (sole >=90 candidate, regular
+file, ambiguous veto) and the read_file_tool wiring: a confident typo is
+retried with the corrected path; ambiguous or failing retries fall back
+to the existing hint-only error contract.
+"""
+
+import json
+import os
+import shutil
+import tempfile
+import unittest
+from unittest.mock import MagicMock, patch
+
+from tools.file_tools import read_file_tool
+from tools.path_validation import confident_nearby_match
+
+
+class _TmpDir(unittest.TestCase):
+    def setUp(self):
+        self._tmp = tempfile.mkdtemp(prefix="test_autoretry_")
+        self.config = os.path.join(self._tmp, "config.yaml")
+        with open(self.config, "w") as f:
+            f.write("key: value\n")
+
+    def tearDown(self):
+        shutil.rmtree(self._tmp, ignore_errors=True)
+
+
+class TestConfidentNearbyMatch(_TmpDir):
+    def test_none_for_existing_path(self):
+        self.assertIsNone(confident_nearby_match(self.config))
+
+    def test_same_stem_typo_corrected(self):
+        self.assertEqual(
+            confident_nearby_match(os.path.join(self._tmp, "config.yam")),
+            self.config,
+        )
+
+    def test_ambiguous_twin_returns_none(self):
+        with open(os.path.join(self._tmp, "config.yml"), "w") as f:
+            f.write("other\n")
+        self.assertIsNone(
+            confident_nearby_match(os.path.join(self._tmp, "config.yam"))
+        )
+
+    def test_weak_match_returns_none(self):
+        self.assertIsNone(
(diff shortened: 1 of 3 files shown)


Lexus2016:
Closing after adversarial code review due to critical security and correctness blockers:

1. **Security bypass on protected files**: The retry path directly invokes 'file_ops.read_file(_fixed)' without re-running '_is_blocked_device', 'has_binary_extension', or 'get_read_block_error'. Requesting a missing typo like 'auth.jso' allows reading protected credentials like 'auth.json' in 'HERMES_HOME', bypassing security boundaries.
2. **Missing redaction and response size caps**: Early return bypasses 'redact_sensitive_text' and content size limits, risking token explosion and secret leaks.
3. **Environment discrepancy**: 'confident_nearby_match()' executes on the local host filesystem rather than the targeted remote/container 'file_ops' environment, causing broken lookups or incorrect host file reads.
4. **Dangerous stem collisions**: Replacing requests like 'config.py' with 'config.yaml' simply because they share a stem produces unintended file reads rather than typo correction.
5. **Missing lifecycle updates**: Bypasses staleness tracking, read history, and deduplication records.

Please revise the proposal to maintain full security checks, environment isolation, and stricter typo constraints.

Issue #942, opened 2026-07-13
[FIX] Terminal retry spiral persists — 55 consecutive retries across 404 sessions despite loop guard


Lexus2016:
## Problem (observed in real usage)

The terminal tool exhibits retry spirals of up to 55 consecutive failed calls across 404 sessions in the last 7 days. Despite existing loop guard infrastructure and prior fixes (#888, #902), the agent continues to retry the same or similar failing terminal commands without breaking out of the cycle. The loop guard either does not fire early enough or the agent finds alternative command variations that bypass the guard.

### Evidence (aggregated, anonymized)
- **Frequency:** 55 max consecutive retries, 404 sessions affected in 7 days
- **Tool:** terminal
- **Failure shape:** agent calls terminal with a command that fails (timeout, blocked, non-zero exit), receives the error, and retries the same or slightly varied command repeatedly instead of switching strategy
- **Secondary impact:** wastes significant token budget and stalls tasks that could proceed via alternative tools (read_file, search_files, write_file)

### Impact on real tasks
The highest-severity retry spiral in the entire tool fleet. 404 sessions means roughly 11% of all sessions experience this pattern. Tasks stall for many turns, and the agent burns budget on commands that will never succeed with the current approach.

## Proposed Solution

1. **Circuit breaker for terminal retries:** After N consecutive failures (configurable, default 5) for the same tool within a session, insert a non-retryable diagnostic message that explicitly tells the agent: "You have failed N consecutive terminal calls. Stop retrying terminal and use an alternative tool (read_file, search_files, write_file) or report the blocker."
2. **Failure-class-aware escalation:** Classify terminal failures (timeout vs blocked vs non-zero exit) and provide class-specific guidance — e.g., "this command requires approval that is not available in cron context" rather than a generic error.
3. **Consecutive-failure cap in loop guard:** Add a hard consecutive-failure count to the existing loop guard that triggers a forced strategy switch, not just a warning.

### Value
- **Impact**: 0.9 (404 sessions, highest-severity spiral)
- **Effort**: 0.4 (extend existing loop guard + error classifier)
- **Priority Score**: 0.9 × 2 × (1 − 0.4 × 0.4) = 1.488

### Research Evidence
- Trace miner output: `terminal` retry spiral up to 55 consecutive across 404 session(s) — 7-day window
- Prior fixes #888 (terminal empty error), #902 (pending_approval loop) addressed specific sub-patterns but not the spiral itself
[...]

PR #2427, opened 2026-08-14
fix(tools): circuit breaker half-open recovery — unstick the read/patch path (Closes #2423)

### LINKED ISSUE #2423: [FIX] Read-tool circuit breaker opens and never half-open probes — whole session loses read/patch path (non-retryable persistent_error)
Read-path tools (file-read, file-search, patch) open a circuit breaker after N consecutive failures — and it then stays open for the REST of the session with no half-open probing. Every further call fails instantly as `Non-retryable: persistent_error`, so a run of recoverable path mistakes permanently removes the entire read path (and patch, which must read its target) from the session.

- Frequency: 4+ independent sessions on 2026-08-14 alone; first-hand reproduced in the introspection session itself — breaker-open errors at 7 consecutive failures (file-read), 12 (file-search), 5 (patch)
- Identical breaker counts (12/7/5) observed across SEPARATE sessions the same day, suggesting breaker state persists across sessions/contexts or is hit identically on first use
- Digest context: read-tool failure rate 5.9%/session, search 2.6%/session — the per-call failures are recoverable, but the breaker converts them into session-wide capability loss
- Tool/area: circuit-breaker implementation over read-path tools

### Impact on real tasks
Once open, the agent cannot read any file for the rest of the session: idempotency guards, dedup against prior reports, and result verification all become impossible. Agents fall back to inferring state from memory notes, degrading every downstream stage that depends on reading artifacts.

### Proposed direction
- Add half-open probing: after a cooldown, allow ONE trial call; success resets the breaker
- Reset breaker state at session/context start; never share breaker state across sessions
- On a "breaker open" error, surface the tool, the failure count, and a concrete recovery hint (wait/probe/use alternative) instead of a bare non-retryable flag

### Value
- Impact: 0.9 (session-wide loss of the read path, multi-session recurrence)
- Effort: 0.4
- Priority Score: 1.51

Lexus2016:
Automated evolution PR for issue #2423.

**Problem.** The per-tool circuit breaker opened after N consecutive failures and stayed open for the rest of the process: the fail-fast gate in `handle_function_call` suppressed the very success that would have reset it. A run of cd8e2305fd ... [snip for brevity in command] ...

### DIFF
--- a/agent/tool_error_recovery.py
+++ b/agent/tool_error_recovery.py
@@ -22,6 +22,7 @@ import enum
 import json
 import logging
 import re
+import time
 from dataclasses import dataclass, field
 from typing import Optional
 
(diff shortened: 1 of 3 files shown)


Lexus2016:
Holding merge on purpose.

This fix is still wanted (#2423 is real). Local tests: 8/8 half-open + Windows footguns clean after merging #2422.

Do **not** squash onto current `origin/main` while the parallel cluster-2/4 upstream sync is in flight. That stack already rewrote `model_tools.py` (~960 lines). Landing this 7-line `handle_function_call` note + breaker change on old `main` would force a second conflict into that sync.

Rebase onto the cluster branch (or onto `main` after the CLI/runtime cluster PR lands), then merge. `agent/tool_error_recovery.py` was not in the cluster-2 file list — that half should apply cleanly.

Lexus2016:
Superseded by #2431 — the half-open recovery is cherry-picked onto the cluster-2/4 stack (cluster 2 had dropped the #942 gate from model_tools.py; #2431 restores the gate and adds half-open). Merging this PR onto current main would conflict with that rewrite.

PR #2623, opened 2026-08-16
fix(process): decompose 'other' error bucket into reason subclasses with recovery hints (Closes #2621)

### LINKED ISSUE #2621: [FIX] process tool 'other' error bucket undecomposed — 90 failures/7d, 18-deep spiral, no recovery hint

The `process` tool's `other` error bucket is the largest undecomposed failure
mode for that tool, and it recurs across many sessions with deep retry spirals.
The only prior `process` issue (#2241) was closed without decomposing this
bucket, so the opaque errors still give the agent no recovery guidance.

- Frequency: 90 failures in the 7-day window across 20 sessions; per-session
  rate ~0.0032 (the tool's dominant failure mode).
- Tool/area involved: `process` (background-process management).
- Failure shape: errors classified as `other` (opaque) — no reason subclass,
  no recovery hint. Retry spirals reach 18 consecutive runs in a single
  session, indicating the agent retries blindly because the error carries no
  actionable signal.

### Impact on real tasks
Background processes (long-running builds, servers, watchers) are a core part
of autonomous work. When `process` fails opaquely, the agent cannot tell
whether the process died, the session was lost, or the poll timed out — so it
retries in a loop, wasting turns and leaving the underlying task unverified.
This is the same fix→regress→refile treadmill seen on other tools: an
undecomposed `other` bucket hides the real cause.

### Proposed direction
Decompose the `process` `other` bucket into reason subclasses (e.g.
session-not-found, process-exited, poll-timeout, invalid-action) with a
targeted recovery hint per subclass, mirroring the decomposition already done
for `tool_call`/`memory`/`tool_describe`. Wire the subclass into the
non-retryable fast-path so a dead session stops being retried.

### Value
- Impact: 0.6 (blocks background-process tasks; moderate frequency, opaque)
- Effort: 0.4
- Priority Score: 0.6 * 2 * (1 - 0.4*0.4) = 1.01

Lexus2016:
Automated evolution PR for issue #2621.

Decomposes the process tool's opaque 'other' error bucket into reason subclasses (session-not-found, invalid-action, session-id-required, process-exited, poll-timeout, other) with a targeted recovery hint per subclass, mirroring the decomposition already done for tool_call/memory/tool_describe. Wires the subclass into the tool_error returns so a dead session stops being blind-retried.

Closes #2621

### DIFF
--- a/tests/tools/test_process_error_decomposition.py
+++ b/tests/tools/test_process_error_decomposition.py
@@ -1,0 +1,97 @@
+"""Tests for process tool error decomposition (#2621).
+
+The ``process`` tool's ``other`` error bucket was the largest undecomposed
+failure mode — opaque errors with no reason subclass and no recovery hint
+drove deep retry spirals. ``_classify_process_error`` maps each error text to
+a concrete category + recovery hint so the agent can act instead of
+blind-retrying, mirroring the decomposition done for ``tool_call``/``memory``/
+``tool_describe``.
+"""
+
+import json
+import pytest
+from unittest.mock import patch
+
+from tools.process_registry import _classify_process_error, _handle_process
+
+
+class TestClassifyProcessError:
+    """Unit tests for the error classifier."""
+
+    def test_session_not_found(self):
+        cls = _classify_process_error("No process with ID 'abc123'")
+        assert cls["error_category"] == "session-not-found"
+        assert "action='list'" in cls["recovery_hint"]
+
+    def test_session_not_found_no_processes(self):
+        cls = _classify_process_error(
+            "No process with ID 'abc123'. No processes are currently registered."
+        )
+        assert cls["error_category"] == "session-not-found"
+
+    def test_invalid_action(self):
+        cls = _classify_process_error("Unknown process action: 'status'")
+        assert cls["error_category"] == "invalid-action"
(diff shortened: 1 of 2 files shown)


Lexus2016:
Closing as DUPLICATE — #2622 (merged today via evolution-integration) already resolves #2621 completely, and this PR was the losing twin in a double-run race (both opened ~4 min apart for the same issue).

Why #2622 won: the issue asks for BOTH (1) decomposition of the process 'other' bucket into reason subclasses with recovery hints AND (2) wiring the subclass into the non-retryable fast-path so a dead session stops being retried. #2622 delivers both — tool_diagnostics rules (consumed by loop_guard.classify) + _NON_RETRYABLE_BY_TOOL['process'] (session_not_found/invalid_action/process_exited), with hard-stop tests. This PR (#2623) delivers the decomposition as tool-side error envelope keys (error_category/recovery_hint, mirroring memory_tool) but does NOT wire the non-retryable fast-path, so the 18-deep blind-retry spiral the issue targets would persist mechanically; and nothing in the codebase consumes the envelope error_category key (verified: only memory_tool emits one; no reader).

The idea is not discarded — it is DONE via #2622. If tool-side envelope hints are still wanted ON TOP of the merged classification, re-open as a small follow-up PR against current main.

Issue #2333, opened 2026-08-12
[FIX] read_file 'other' error bucket (217/7d) — undecomposed, 20-deep spiral, distinct from fixed file-not-found


Lexus2016:
## Problem (observed in real usage)
`read_file` has a large undecomposed `other` error bucket that produces opaque failures and 20-deep retry spirals, distinct from the already-fixed `file-not-found` mode.

### Evidence (aggregated, anonymized)
- Frequency: 217 `other` failures in the last 7 days across 27,321 sessions (rate 0.0079/session); 20-deep consecutive retry spiral observed.
- Tool/area involved: `read_file`
- Failure shape: `read_file` returns an unclassified `other` error (not file-not-found, not permission-denied) with no recovery hint; agent retries the same read repeatedly.

### Impact on real tasks
Reading files is the most common file operation. Opaque `other` failures stall the agent mid-task and the retry spiral wastes turns. Prior fixes (#2242, #2293) addressed the `file-not-found` mode specifically; the `other` bucket was never decomposed and remains the second-largest `read_file` failure mode.

### Proposed direction
Decompose `read_file` `other` into concrete subclasses (encoding, binary, path-type, size-limit) with a targeted hint per subclass, and add a non-retryable hard stop for the unclassified remainder — mirroring the reason-scoped approach that successfully reduced `file-not-found`.

### Value
- Impact: 0.6 (frequent, blocks file reads)
- Effort: 0.4
- Priority Score: 0.6*2*(1-0.4*0.4) = 1.01

Lexus2016:
PR #2337 was closed in code-review — it bundled terminal-timeout changes (now redundant: #2335 closed by merged PR #2338), read_file 'other' decomposition (#2333), and patch 'ambiguous' strengthening (#2334) into one PR that fails CI on test slice 6/8.

To fix: file a SMALL, focused PR for JUST this issue's scope (read_file error decomposition / patch ambiguous wording). Do NOT bundle with terminal-timeout changes. The classifier test suite is sensitive to timeout-retry behavior changes — keep this PR's changes to file_operations.py error classification only.

PR #2722, opened 2026-08-17
feat: decompose terminal 'other' error bucket into reason subclasses (Closes #2717)

### LINKED ISSUE #2717: [FIX] terminal 'other' error bucket undecomposed — 396 failures/7d, largest residual, no recovery hint

The `terminal` tool's opaque `'other'` error bucket is the largest undecomposed
failure mode in the agent's tool surface. Every sibling tool's `'other'` bucket
has already been decomposed into actionable reason subclasses (read_file #2333,
memory #2332, tool_call #2336, patch #2244, process #2621, vision_analyze #2310,
tool_describe #2309) — but `terminal`'s residual `'other'` bucket was never
split, so it remains the biggest single source of unactionable errors.

- Frequency: 396 `terminal` failures classified `'other'` in the last 7 days
  (28,536 sessions scanned), at a per-session rate of ~0.0139.
- Tool/area involved: `terminal` (the most-used tool in the agent).
- Failure shape: an unclassified `terminal` error with no reason subclass and no
  recovery hint, so the agent cannot tell whether to retry, change strategy, or
  stop.

### Impact on real tasks
`terminal` is the workhorse tool. When it fails with an opaque `'other'` error,
the agent has no structured guidance to recover — it either retries blindly
(contributing to the recurring terminal retry-spiral problem) or stalls. This
directly blocks autonomous task completion on the tool used most often.

### Proposed direction
Decompose the `terminal` `'other'` bucket into reason subclasses (e.g.
environment/executable-not-found, shell-parse, output-encoding, resource-limit,
unknown) with per-subclass recovery hints, mirroring the pattern already applied
to the sibling tools. Route the residual through a shared classifier so it never
surfaces as opaque `'other'`.

### Value
- Impact: 0.8
- Effort: 0.3
- Priority Score: 1.41

Lexus2016:
Automated evolution PR for issue #2717.

Decomposes the terminal tool's opaque 'other' error bucket into actionable reason subclasses (env_backend_error, shell_parse_error, encoding_error, resource_limit), mirroring the pattern already applied to sibling tools (#2621 process, #2333 read_file, etc.). Each subclass carries a targeted recovery hint so the agent can tell whether to retry, change strategy, or stop.

Rules are ordered most-specific-first and run before the generic runtime_error catch-all. The env_backend_error rule is placed before the timeout rule (which matches 'connection refused' / 'SSH connection failed').

### DIFF
--- a/agent/tool_diagnostics.py
+++ b/agent/tool_diagnostics.py
@@ -137,6 +137,40 @@ _RULES: tuple[tuple[re.Pattern, str, str], ...] = (
         "or a different API). If no alternative exists, escalate to the "
         "user with the exact access needed. Do NOT retry the same action.",
     ),
+    # #2717 — terminal "other" bucket decomposition. Terminal is the most-used
+    # tool; its residual 'other' bucket (396 failures/7d, ~0.0139 per-session)
+    # was the largest undecomposed failure source — every sibling tool's
+    # 'other' bucket had already been split (#2333/#2332/#2336/#2244/#2621/
+    # #2310/#2309) but terminal's hadn't. These rules decompose terminal-
+    # specific failure shapes that the exit-code family (#2306) and
+    # missing_command rules above DON'T catch — environment/backend failures,
+    # encoding issues, resource limits, and shell-parse errors — into reason
+    # subclasses with targeted recovery hints BEFORE the runtime_error catch-all.
+    # Ordered most-specific first; must run BEFORE the catch-all (which matches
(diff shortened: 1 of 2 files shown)


Lexus2016:
Closing as a duplicate — PR #2721 (created earlier, 20:30Z) already implements #2717 and is green + within the 200-line merge cap. This PR (#2722) was opened in a later run without first checking for an existing open PR for the same issue. Deferring to #2721 to keep the funnel clean.

PR #2836, opened 2026-08-18
fix(cron): env-exec capability gate for direct stage dispatch (Closes #2826)

### LINKED ISSUE #2826: [FIX] Subagent env still lacks terminal/shell after preflight #2758 — implementation/integration stages cannot execute git/gh

Lexus2016:
Automated evolution PR for issue #2826.

## Problem
Implementation/integration stage crons (`evolution-implementation`, `evolution-integration`) run directly through the scheduler and bypass the Hydra gate's P-001 env-capability check. When the environment lacks execution tooling (no terminal/shell/git), a stage is dispatched and immediately returns BLOCKED — wasting an orchestrator tick and writing no artifact.

The preflight (#2758) detects the missing capability for the Hydra orchestrator only; the direct stage crons were never covered.

## Fix
Extend the fail-fast P-001 exec-capability gate to the scheduler's direct stage dispatch:
- `cron/evolution_preflight.py`: add `stage_requires_exec()` (implementation/integration require execution; introspection/analysis are read-only) and `_check_exec_capability()` (mirrors Hydra gate: git on PATH; fail-open on probe error).
- `cron/scheduler.py`: after the halt-state gate, skip exec-required stages with a silent doc when capability is absent.
- Tests: `TestExecCapabilityGate` covers stage classification + capability probe paths.

## Checks
- Pre-PR test runner: PASSED (tests/cron/test_scheduler.py shard)
- `pytest tests/cron/test_evolution_preflight.py`: 48 passed
- `ruff check`: All checks passed
- Diff: 152 lines (within 200-line self-merge cap)

Closes #2826

### DIFF
--- a/cron/evolution_preflight.py
+++ b/cron/evolution_preflight.py
@@ -12,9 +12,10 @@ from __future__ import annotations
 
 import logging
 import os
+import subprocess
 import time
 from pathlib import Path
-from typing import Any, Dict, Optional
+from typing import Any, Dict, Optional, Tuple
 
 from hermes_constants import get_hermes_home
 from hermes_cli.config import load_config_readonly
@@ -43,6 +44,18 @@ _HALT_GATED_STAGES = frozenset({
     "analysis",
     "implementation",
     "research",
+})
+
+# Stages whose deliverable REQUIRES a real execution environment (git/gh/
+# tests) — the subagent must be able to run commands, not just read/write
+# files. If the environment lacks the execution tooling (no git on PATH),
+# dispatching these stages is doomed: the subagent returns BLOCKED and writes
+# no artifact. `implementation` writes code+PRs (needs git/gh/ruff/pytest);
+# `integration` merges PRs (needs git/gh/hermes update). Introspection and
+# analysis are read-only (file/web) and never exec-gated (#2826).
+_EXEC_REQUIRED_STAGES = frozenset({
+    "implementation",
+    "integration",
 })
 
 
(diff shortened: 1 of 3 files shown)


Da-Mikey:
**Evolution integration — review passed, CI-blocked (repo-wide Nix break).**

Code review: ✅ PASSED. Extends the P-001 exec-capability fail-fast to direct stage dispatch: verified `stage_requires_exec` + `_check_exec_capability` exist in `cron/evolution_preflight.py`, the gate is wired in `_run_job_impl` right after the halt-state gate with `stage`/job names in scope, fail-open on probe error, silent-skip doc on block (`SILENT_MARKER`); tests cover stage classification (implementation/integration vs read-only) and all three probe paths. Matches issue #2826's intent.

CI: all branch-protection REQUIRED checks green (Python tests 12/12 slices, lints, security scans). The `Nix` workflow is red **repo-wide** — flake evaluation error (`bubblewrap-0.11.2 … not available on the requested hostPlatform`) failing identically on `main` and every open PR; this PR does not touch nix files and the re-run (attempt 2) failed identically, confirming it is deterministic and repo-level. Not a PR defect → no rework; left open for the next integration cycle / owner discretion. (Note: PR #2829 also claims `Closes #2826` and was sent back for rework today — this PR is the green candidate for that issue.)

Lexus2016:
Blocked by code review — do not merge as-is. Category error (details on issue #2826): `_check_exec_capability()` probes the PARENT scheduler env for git on PATH (copied verbatim from the hydra gate #2758 probe), but this issue's evidence is that SUBAGENTS are provisioned without shell/terminal while the parent has tooling — so the gate cannot fire in the reported scenario and the doomed dispatch proceeds. Also the skip returns silent success, hiding the environment defect from digests. Rework brief on #2826; re-open a PR once fixed.

PR #2843, opened 2026-08-19
fix(cron): per-job exponential backoff + auto-pause circuit breaker (Closes #70)

### LINKED ISSUE #70: [FIX] Web scraping and RSS fetching frequently blocked by HTTP 403 / Access Denied without retry or fallback

### Impact on real tasks

### Proposed direction

### Value
- Impact: 0.60
- Effort: 0.40
- Priority Score: 3.00

Da-Mikey:
Automated evolution PR for issue #70 (Da-Mikey/hermes-agent-evolution).

## What
Per-job failure accounting in `_mark_job_run_locked`: byte-identical consecutive errors accumulate `failure_streak`; backoff skips `2**(streak-1)-1` scheduled slots (cap `cron.failure_backoff_max_skips` default 16); a success or changed error resets the streak; at `cron.failure_auto_pause_threshold` (default 10) the job auto-pauses (enabled=False, state=paused, paused_reason) as a circuit breaker. Manual resume clears the streak. Adds `_advance_schedule_by_steps` for interval and cron schedules.

## Why
Introspection finding `cron_deterministic_400_burst_no_backoff`: cron job 0b013aa0bd43 failed 4 consecutive times on a byte-identical Gemini-400, re-running a deterministic failure on the fixed schedule and burning tokens/disk.

## Tests
- 9 new tests in `tests/cron/test_jobs.py::TestFailureBackoff` (streak accounting, backoff skips, changed-error reset, success reset, auto-pause at threshold, resume clears streak, advance-by-steps interval/zero, config defaults).
- Full cron suite: 879 passed, 4 skipped.
- `ruff check` clean.

## Notes
This was previously mis-triaged as "already-exists" because the analysis stage read the uncommitted working-tree copy of `cron/jobs.py` (which already contained this implementation) and attributed it to `canonical/main`. Verified this session: no `failure_backoff` symbol exists on `lexus/main`, `origin/main`, or `canonical/main` — the code existed only uncommitted.

Closes #70.

### DIFF
--- a/cron/jobs.py
+++ b/cron/jobs.py
@@ -964,6 +964,63 @@ def compute_next_run(schedule: Dict[str, Any], last_run_at: Optional[str] = None
     return None
 
 
+def _failure_backoff_config() -> Dict[str, Any]:
+    """Return the ``cron`` config section for failure-backoff tuning (#70).
+
+    Read-only: uses ``load_config_readonly`` so the hot path (every job run)
+    does not pay the defensive deepcopy. A missing config yields an empty
+    dict so all call sites fall back to their built-in defaults.
+    """
+    try:
+        from hermes_cli.config import load_config_readonly
+        cfg = load_config_readonly() or {}
+        cron_cfg = cfg.get("cron") if isinstance(cfg, dict) else None
+        return cron_cfg if isinstance(cron_cfg, dict) else {}
+    except Exception:
+        return {}
+
+
+def _advance_schedule_by_steps(
+    schedule: Dict[str, Any],
+    from_iso: str,
+    n_steps: int,
+) -> Optional[str]:
+    """Advance ``from_iso`` forward by ``n_steps`` schedule slots.
+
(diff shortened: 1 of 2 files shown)


Da-Mikey:
Code review FAILED (evolution integration, 2026-08-19) — sending back for rework.

Blocking defect: unguarded int() on user config. `_failure_backoff_config().get('failure_backoff_max_skips', 16)` (and `failure_auto_pause_threshold`) is passed to int() with no coercion guard; a non-int value in config.yaml raises inside `_mark_job_run_locked` on the FAILURE path, aborting before save_jobs — last_status/last_error are lost and the run path throws (empirically reproduced). Also: synthetic interruption/forced-release failures (scheduler shutdown-interrupt, forced-release marks) count toward failure_streak and can false-positive auto-pause a healthy job; and a stale streak survives re-enables that don't go through resume_job, re-tripping the breaker immediately. Rework brief posted on issue #70.

PR #2856, opened 2026-08-19
fix: cron scheduler backoff/circuit breaker (Closes #70)

### LINKED ISSUE #70: [FIX] Web scraping and RSS fetching frequently blocked by HTTP 403 / Access Denied without retry or fallback

Da-Mikey:
Automated evolution PR for issue 70-cron-failure-backoff-r2. Rework of a prior closed PR per the issue's review brief; validated locally (ruff check clean, targeted tests green) before opening.

### DIFF
--- a/cron/jobs.py
+++ b/cron/jobs.py
@@ -964,6 +964,97 @@ def compute_next_run(schedule: Dict[str, Any], last_run_at: Optional[str] = None
     return None
 
 
+def _failure_backoff_config() -> Dict[str, Any]:
+    """Return the ``cron`` config section for failure-backoff tuning (#70).
+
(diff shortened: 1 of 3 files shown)


Da-Mikey:
Closing per evolution code review: CI slice 11/12 fails deterministically — `tests/cron/test_shutdown_interrupt.py` mock `mark()` has no `status` kwarg, so `mark_running_jobs_interrupted("shutdown")` marks nothing (assert [] == ['same-job','same-job']). Rework brief posted on Da-Mikey/hermes-agent-evolution#70.

PR #2902, opened 2026-08-19
fix: BLOCKED (hardline) names the offending command and is non-retryable (Closes #2873)

### LINKED ISSUE #2873: [FIX] terminal 'BLOCKED (hardline): command parser limit or malformed executable payload' gives no actionable signal — agent retries identical blocked commands (3x in one subagent session 20260819_004618_b8fc0a)
## Source
Introspection 2026-08-19: 11 'BLOCKED (hardline)' / blocklist events in errors.log within the 7-day window; 3+ in session 20260819_004618_b8fc0a on 08-19 (re-attempts at 00:49:20, 00:50:35, 00:51:33). No existing issues/*.json artifact mentions hardline/blocklist; live gh open-issue set has no such issue.

## Summary
The terminal error 'BLOCKED (hardline): command parser limit or malformed executable payload' carries no actionable signal: it does not name the offending command or the matching blocklist pattern, so the agent cannot tell whether the command itself, its payload, or an unrelated parser limit caused the block. Recovery is blind — observed behavior is identical re-attempts (3x in one session).

## For Hermes
The same session also shows exit_code=1 with exit_code_meaning 'No matches found (not an error)' counted as a failure — the trace-miner extractor counts any exit_code!=0 as failure; a classification nuance worth separating in the same fix.

## Proposed action
(1) Include the offending command/pattern (redacted) and the block reason in the BLOCKED error text; (2) treat BLOCKED as non-retryable-with-same-command in the retry/breaker layer (mirror #2228's non-retryable diagnostic), forcing the agent to change approach; (3) separate exit_code=1 'no matches' from real failures in trace-miner classification.

## Priority
0.88 (impact 0.5, effort 0.3).

## Related
#2228 (non-retryable diagnostic), #1496/#2302/#2362 (retry-spiral family — distinct surface: this is error-signal actionability, not retry policy), #2758 (preflight/hardening family).

Lexus2016:
Automated evolution PR for issue #2873.

The terminal 'BLOCKED (hardline)' error carried no actionable signal — it named the block category but not the offending command, so the agent blindly retried identical blocked commands (observed 3x in one subagent session).

Changes:
1. `tools/approval.py` — `_hardline_block_result` now includes a redacted 'Offending command' preview (truncated at 300 chars, secrets masked via redact_sensitive_text), an explicit 'Do NOT retry' directive, and a machine-readable `retryable: False` marker so the retry/breaker layer does not re-run the same payload.
2. `scripts/introspection_extract.py` — `_tool_result_failed` no longer counts annotated informational exits (grep/rg/diff exit_code=1 'No matches found (not an error)') as failures, so trace-miner classification separates expected no-match results from real failures.

Validation: ruff check ✓, targeted tests ✓ (73 passed incl. 4 new #2873 tests).

### DIFF
--- a/scripts/introspection_extract.py
+++ b/scripts/introspection_extract.py
@@ -93,9 +93,21 @@ def _tool_result_failed(content: Any) -> bool:
         return False
     if "exit_code" in data:
         try:
-            return int(data["exit_code"]) != 0
+            code = int(data["exit_code"])
         except (TypeError, ValueError):
             return False
+        if code == 0:
+            return False
+        # #2873: the terminal tool annotates informational non-zero exits
+        # (grep/rg/diff "no matches", test false) with
+        # exit_code_meaning="... (not an error)". Those are NOT failures —
+        # counting them as failures polluted the trace-miner signal and sent
+        # the agent re-investigating expected exit codes.
+        if code == 1:
+            meaning = str(data.get("exit_code_meaning") or "").lower()
+            if "not an error" in meaning:
+                return False
+        return True
     if data.get("error") or str(data.get("status", "")).lower() == "error":
         return True
     for ok_key in ("success", "ok"):
--- a/tests/tools/test_blocked_command_guidance.py
+++ b/tests/tools/test_blocked_command_guidance.py
@@ -49,6 +49,36 @@ class TestParserLimitRecovery:
         # And nothing was saved for a genuine hardline block.
         assert not (tmp_path / ".hermes" / "cache" / "blocked-scripts").exists()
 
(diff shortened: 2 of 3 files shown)


Lexus2016:
Superseded by #2914 (same issue #2873, opened 21 min later). #2914 covers everything #2902 does — actionable hardline signal in tools/approval.py + introspection_extract.py exit_code_meaning handling — and adds the explicit non-retryable rule in tools/tool_failure_classifier.py (the classifier-layer fix #2902 lacks). Keeping the issue open; it will close via #2914's Closes.

PR #2905, opened 2026-08-19
fix: delegation gating rubric — compress COST GATE prose (Closes #69)

### LINKED ISSUE #69: [FIX] Agent repeatedly fails to install Python packages in PEP 668 externally-managed environments

Da-Mikey:
Automated evolution PR for issue #69 (r4 rework).

Prior PRs (#2880) blew the `_build_top_level_description()` compaction contract (2816 chars vs ≤2200). This rework compresses the when-to-delegate COST GATE + LAST-KNOWN STATE guidance into the static description (now 2129 chars) while keeping every rubric keyword and all pre-existing USE FOR / DO NOT USE FOR routing needles.

Local validation: `pytest tests/tools/test_delegate.py tests/agent/test_delegate_gating_rubric.py` → 193 passed; `ruff check` clean.


Lexus2016:
Closing as stale: this PR's Closes target, issue #69, is already CLOSED — it can never pass the integration gate ('closes a real, open issue'). If the delegation-gating rubric prose change is still wanted, re-file against current main.

PR #2927, opened 2026-08-19
fix(cron): backoff + circuit breaker for deterministic failures — r3 rework (Closes #70)

### LINKED ISSUE #70: [FIX] Web scraping and RSS fetching frequently blocked by HTTP 403 / Access Denied without retry or fallback

Da-Mikey:
## Summary

Per-job exponential backoff + a circuit breaker so deterministically failing scheduled jobs are no longer retried on their fixed schedules indefinitely.

This is the **r3 rework** addressing the review briefs on the two prior closed PRs (#2843, #2856).

## What changed vs. the earlier attempts

1. **Unguarded `int()` on config fixed** — `cron.failure_backoff_max_skips` / `cron.failure_auto_pause_threshold` are now coerced via try/except with fallback to defaults (16 / 10). A non-int config.yaml value no longer crashes the failure path of `_mark_job_run_locked`.
2. **Synthetic failures excluded from streak accounting** — `status="interrupted"` / `status="released"` marks (scheduler restart / forced release) never advance `failure_streak`, so restarts can't false-positive auto-pause a healthy job.
3. **Streak cleared on every re-enable path** — `resume_job`, `trigger_job`, `update_job` (enabled transition), and forced fire claim all reset `failure_streak` / `failure_backoff_skips`.
4. **Backward-compatible `status` kwarg** — the scheduler call sites and the shutdown-interrupt test double now accept `status`, fixing the deterministic CI slice failure from PR #2856.

## Verification

- `pytest tests/cron/ -q` → **889 passed, 4 skipped**
- `ruff check cron/jobs.py cron/scheduler.py tests/cron/test_jobs.py tests/cron/test_shutdown_interrupt.py` → clean
- 15 new tests covering: single/identical/changed failures, backoff caps, auto-pause threshold, non-int config fallback (both knobs), interruption-not-counted, resume/trigger/update/forced-fire clearing the streak.

Closes #70.

### DIFF
--- a/cron/jobs.py
+++ b/cron/jobs.py
@@ -964,6 +964,97 @@ def compute_next_run(schedule: Dict[str, Any], last_run_at: Optional[str] = None
     return None
 
 
+def _failure_backoff_config() -> Dict[str, Any]:
+    """Return the ``cron`` config section for failure-backoff tuning (#70).
+
+    Read-only: uses ``load_config_readonly`` so the hot path (every job run)
+    does not pay the defensive deepcopy. A missing config yields an empty
+    dict so all call sites fall back to their built-in defaults.
+    """
+    try:
+        from hermes_cli.config import load_config_readonly
+
+        cfg = load_config_readonly() or {}
+        cron_cfg = cfg.get("cron") if isinstance(cfg, dict) else None
+        return cron_cfg if isinstance(cron_cfg, dict) else {}
+    except Exception:
+        return {}
+
+
+# #70: last_status values recording a SYNTHETIC failure — scheduler restart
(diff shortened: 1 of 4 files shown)


Lexus2016:
Superseded by #2955 — this PR's head lives on the Da-Mikey fork, so the post-rework repairs (main re-merge + the unified failure_streak fix for the double-increment scar) cannot be pushed here. #2955 carries the r3 rework plus both.

PR #2959, opened 2026-08-19
fix: coerce_tool_args JSON-parse-failure warning includes tool.param (Closes #2953)

### LINKED ISSUE #2953: [FIX] coerce_tool_args JSON-parse-failure path passes uncoerced string to tools — 2.4k opaque WARNINGs/5d, 23% followed by tool errors
The generic argument-coercion path for **list-typed tool parameters** still fails at scale: when the model emits a bare string for a list-typed argument and JSON parsing fails, the runtime logs a WARNING that (a) does not name the tool/parameter, and (b) passes the **uncoerced string** through to the tool instead of falling back to a single-element wrap. This leaves the tool with a malformed argument and gives the agent no recovery signal.

- Frequency: **2,381 occurrences in the last 5 days** (~450–635/day, growing: 188 → 635 → 579 → 529 → 450), observed across many sessions, not a single incident.
- Tool/area involved: `model_tools.coerce_tool_args` (all tools with array-typed parameters).
- Failure shape: warning text `failed to parse string as JSON for expected type list: Expecting value: line 1 column 1 (char 0)` — the value is not JSON-shaped, parse fails, the original string is returned and dispatched to the tool.
- Correlation: **23% of these warnings are followed by a tool error within 5 log lines** in the same session (e.g. patch no-match, path-not-found) — the malformed argument reaches the tool and contributes to downstream failure buckets.
- Log impact: these lines are ~23% of the errors.log volume, polluting the failure signal the pipeline digests.

### Impact on real tasks
Every occurrence is a model→tool argument contract violation that the agent cannot see or recover from (no tool/param name, no hint). At 500+/day this is a persistent source of avoidable tool failures, retries, and log noise that obscures genuine defects.

### Proposed direction
In the JSON-parse-failure branch for list-typed params, mirror the existing single-element fallback already used by the sibling path (wrap the string as `[value]`) so the tool still receives a valid argument, and include `tool_name.param` + the offending value prefix in the WARNING so the agent gets a deterministic recovery hint (same family as the earlier read_file stringified-array guard, extended to the generic pure-list path). Add a regression test asserting the fallback on a non-JSON string for a list param.

### Value
- Impact: 0.7
- Effort: 0.3
- Priority Score: 1.23

Lexus2016:
Automated evolution PR for issue #2953. Threads tool_name.key through _coerce_value to _coerce_json so the list-failure warning names the tool/param. Single-element wrap already handled at call site; no double-wrap.

### DIFF
--- a/model_tools.py
+++ b/model_tools.py
@@ -848,7 +848,9 @@ def coerce_tool_args(tool_name: str, args: Dict[str, Any]) -> Dict[str, Any]:
         # (e.g. read_file's normalize_read_pagination) already handle it.
         if expected == "array" and value is not None and not isinstance(value, (list, tuple)):
             if isinstance(value, str):
-                coerced = _coerce_value(value, expected, schema=prop_schema)
+                coerced = _coerce_value(
+                    value, expected, schema=prop_schema, tool_name=tool_name, key=key
+                )
                 if coerced is not value:
                     # _coerce_value handled it (JSON-parsed list or
                     # nullable "null" → None).
@@ -932,7 +934,9 @@ def coerce_tool_args(tool_name: str, args: Dict[str, Any]) -> Dict[str, Any]:
             continue
         if not expected and not _schema_allows_null(prop_schema):
             continue
-        coerced = _coerce_value(value, expected, schema=prop_schema)
+        coerced = _coerce_value(
+            value, expected, schema=prop_schema, tool_name=tool_name, key=key
+        )
         if coerced is not value:
             args[key] = coerced
             # If we just JSON-parsed a string into a container, recurse so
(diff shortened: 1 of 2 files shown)


Da-Mikey:
Heads-up for the owner (evolution integration pipeline): this PR and #2960 both target issue #2953. #2960 is the superset — it includes the tool.param naming in the JSON-parse-failure warning (this PR's change) *plus* the bare-string wrap for list-typed union params. If #2960 merges, this PR is redundant; consider closing it. (Both are currently blocked by the known systemic nix workflow failure — the Diagnose step `nix run .#fix-lockfiles` crashes on both legs, unrelated to this diff.)

Lexus2016:
Superseded by #2960 (same issue #2953, opened 1 min later). #2960 contains this PR's warning-context change (tool.param named in the parse-failure WARNING) PLUS the actual fix for the reported failure mode — the wrap fallback for bare strings on array-accepting non-string schemas (`_schema_accepts_kind` path for union ["array", "null"]). This PR only threaded tool/key into warnings and would conflict with #2960 on the same function signatures; the complete fix is in #2960. Keeping #2953 accepted — it closes when #2960 lands.

PR #2990, opened 2026-08-20
feat(memory): contradiction detection for episodic memory (Closes #37)

Da-Mikey:
Automated evolution PR for issue #37 — bounded first slice of the 4-tier memory management work: contradiction handling on the episodic write path.

When a new observation conflicts with a stored memory (negation flip on a shared subject), both sides are surfaced with timestamps instead of being silently overwritten — the issue's explicit success criterion.

- `agent/memory_contradiction.py` (stdlib-only): deterministic negation-flip detector. Requires >=2 shared content tokens (subject + claim) AND a polarity mismatch, so a shared predicate alone never false-positives. Returns ContradictionFlag (both texts + both timestamps).
- `agent/memory_manager.py`: `MemoryManager.check_contradictions()` runs the detector over the recent episodic store (bounded window, importance filter) and logs a warning per flag; `score_memories()` calls it on every scored turn — non-fatal by design.
- 15 unit + wiring tests; ruff check/format + footgun scan clean; existing memory_importance tests unaffected.

Remaining slices (semantic-tier staleness, consolidation cron, working-memory eviction) are separate cycles — issue stays open.

### DIFF
--- a/agent/memory_contradiction.py
+++ b/agent/memory_contradiction.py
@@ -1,0 +1,303 @@
+# -*- coding: utf-8 -*-
+"""Contradiction detection for episodic memory (issue #37).
+
+The memory-management research behind #37 identifies contradiction handling as
+a core write-path operation: when a new observation conflicts with a stored
+memory, the system must *surface* both (with timestamps) rather than silently
+overwrite or accumulate noise. This module implements that operation for the
+episodic tier (``agent.memory_importance.EpisodicMemoryStore``) with
+deterministic, LLM-free heuristics:
+
+1. **Negation flip on a shared subject** — if a new observation asserts
+   ``<subject> is X`` while a stored event asserts ``<subject> is not X``
+   (or vice versa), that is a contradiction. Detected by comparing negation
+   polarity between the two texts when they share at least one content token
+   (the subject).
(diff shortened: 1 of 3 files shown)


Da-Mikey:
✅ Ready for merge — evolution pipeline (code reviewed, all required checks pass).

**Review (evolution integration):** `agent/memory_contradiction.py` is wired into a real, reachable call site: `MemoryManager.score_memories` calls `self.check_contradictions(user_content)` on the per-turn episodic write path (non-fatal, bounded window of 200, importance filter). The detector is deliberately conservative (≥2 shared content tokens + polarity mismatch; stopword/negation-fragment exclusions), carries both sides verbatim with timestamps per #37 success criterion, and never raises on malformed store entries. 15 unit + wiring tests incl. contraction negation and no-false-positive cases. Mechanism is deterministic heuristics — correct for this slice; no category errors.

**CI note:** the two `nix` checks are red on this PR for the known repo-wide infra reason (flake never defined `fix-lockfiles`; see #2973, fix in flight in #2975). This PR touches zero nix files. Branch was synced with main during review; fresh CI confirms all non-nix checks green.

PR #3059, opened 2026-08-21
fix(delegation): expand unexpanded template markers at dispatch time (Closes #95)

Da-Mikey:
## Fix: unexpanded template markers silently no-op delegated cron stages

The cron orchestrator (and other LLM-driven dispatchers) copy prompt placeholders (`<NOW-ISO>`, `<session_id>`, `<generated-timestamp>`) verbatim into `delegate_task` goals. `delegate_task`'s batch quality gate then rejected the call, and the delegated stage silently no-op'd every ~30 minutes across days — invisible unless someone read the logs.

**Changes:**

1. **Substitute at dispatch time** — `expand_template_markers()` substitutes the mechanically-resolvable markers (timestamps, session id) with real values inside `delegate_task`, so the common case just works and never trips the rejection guard.
2. **Loud residual rejection** — unresolvable markers (`<real citation>`, `<feature_name>`) are still rejected (the #81141 protection is preserved), but now with a `logger.warning` so the failure surfaces instead of silently skipping the stage.

New `tests/tools/test_delegate_template_markers.py` covers substitution, residual reporting, and the dispatch-integration path (a batch carrying `<NOW-ISO>` now proceeds).

Automated evolution PR for issue #95.

### DIFF
--- a/tests/tools/test_delegate_template_markers.py
+++ b/tests/tools/test_delegate_template_markers.py
@@ -1,0 +1,180 @@
+"""Tests for template-marker expansion at delegate_task dispatch (issue #95).
+
+The cron orchestrator (and other LLM-driven dispatchers) can copy prompt
+placeholders (<NOW-ISO>, <session_id>, <generated-timestamp>) verbatim into a
+delegate_task goal. delegate_task's batch quality gate then rejects the call
+and the delegated stage silently no-ops. These tests prove the fix:
+
+1. Known markers are substituted with real values at dispatch time.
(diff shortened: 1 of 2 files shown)


Da-Mikey:
Closing per evolution integration code review — deterministic dead-code gate (skill 2a): `strip_residual_template_markers` (tools/delegate_tool.py:4776) has zero call sites anywhere in the repo outside its own file and tests — it would ship as unreachable code. The core mechanism (`expand_template_markers` wired into `delegate_task` + loud residual rejection) is correct and reviewed. Rework brief posted on issue #95.

PR #3103, opened 2026-08-22
fix(nix): refresh npm lockfile hashes for CI stability

Lexus2016:
## Summary
Refreshes npm lockfile hashes to fix the Nix flake CI build failures and eliminate false failure notifications.


Lexus2016:
Closing: the Nix build failure is caused by a missing coreutils/cmp binary at the expected Nix store path — this is an upstream nixpkgs/flake.lock staleness issue, not a lockfile hash issue. The fix-lockfiles script itself uses a hardcoded coreutils path that doesn't exist in the current Nix store. This needs an upstream flake.lock update (nix flake update) rather than a package-lock.json refresh.

PR #3172, opened 2026-08-24
feat: memory-primed subagent spawning via opt-in memory briefing (Closes #105)

### LINKED ISSUE #105: [CAPABILITY] Cron agent sessions do not survive gateway restarts — interrupted jobs vanish without record or retry

Da-Mikey:
Automated evolution PR for issue #105.

**What:** Optional `memory_briefing` flag on `delegate_task` that primes spawned children with the parent's long-term memory. When opted in, a bounded, most-relevant-first briefing is assembled via the parent's existing prefetch path (`MemoryManager.prefetch_all`) for the task's goals and prepended to each task's `context` as background reference.

**Why:** Spawned subagents start cold — they receive only the explicit `context` (plus optional collapsed parent summary). Domain-knowledge-heavy delegations produce weaker results when the child must re-derive background the parent already has in long-term memory (research: memory-primed agents outscore cold agents 9.6 vs 7.2 on domain tasks).

**Design:**
- Off by default (`memory_briefing` unset/falsy) — behavior byte-identical to today; the helpers are never invoked.
- Briefing built ONCE per delegate call (fused goal+context query, capped at 2000 chars) — no per-task retrieval multiplication.
- Reuses the existing prefetch path (#75 dosage work) — no new retrieval machinery, no new dependency.
- Injection guard: briefing prepended under a header marking the content as UNTRUSTED DATA (memory-store content is data, not instructions), mirroring the child system prompt's untrusted-content banner.
- Bounded: body hard-capped at 4000 chars, head kept (most-relevant-first per provider), truncation noted.
- Best-effort: no memory manager / empty query / prefetch failure / prefetch raise -> no-op, context untouched.

**Tests:** tests/tools/test_memory_briefing.py — 13 tests pinning: query fusion + cap, no-manager/empty-query/failure no-ops, untrusted-data guard, bounding, prepend semantics preserving explicit context, and the flag gate. All green locally (29/29 together with the existing handoff-collapse tests; ruff check clean).

### DIFF
--- a/tests/scripts/test_evolution_trace_replay.py
+++ b/tests/scripts/test_evolution_trace_replay.py
@@ -56,8 +56,7 @@ class TestIterTrajectories:
         assert [e.tool for e in logs[0].entries] == ["tool_0", "tool_1"]
 
     def test_jsonl_append_format_multi_turn(self, tmp_path):
-        p = tmp_path / "2026-08-23_s1.jsonl"
-        _make_log("success", session_id="s1").append(tmp_path)
+        p = _make_log("success", session_id="s1").append(tmp_path)
         _make_log("failure", session_id="s1").append(tmp_path)
         logs = iter_trajectories(p)
         assert len(logs) == 2
--- a/tests/tools/test_memory_briefing.py
+++ b/tests/tools/test_memory_briefing.py
@@ -1,0 +1,193 @@
+"""Tests for memory-primed spawning (GitHub issue #105).
+
+Spawned children start cold: they receive only the explicit ``context`` string
+(plus, optionally, a collapsed parent-conversation summary). The OPTIONAL
+``memory_briefing`` flag primes them with the parent's long-term memory: a
+bounded, most-relevant-first briefing is assembled via the parent's EXISTING
+prefetch path (``MemoryManager.prefetch_all``) and prepended to each task's
+``context`` as background reference.
+
+These tests pin four contracts:
+
+  1. No memory manager on the parent      -> briefing is None, context untouched.
+  2. Opted in with a working prefetch     -> briefing block prepended to every
+                                             task's context, bounded and marked
+                                             UNTRUSTED DATA (injection guard).
+  3. Prefetch failure / empty result      -> no-op, context unchanged (best-effort).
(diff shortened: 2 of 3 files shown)


Da-Mikey:

Da-Mikey:

Lexus2016:
Resolved via cherry-pick to `main` in commit 4e8f6a0c4a with authorship preserved. Thank you @Da-Mikey!

PR #3205, opened 2026-08-26
feat(audit): per-call subagent attribution stamping in audit trail

Lexus2016:
Implements local proposal evo-2026-08-26-03 (no GitHub issue — local_only proposal).

## What
Propagate the existing `HERMES-SUBAGENT-ATTRIBUTION` marker (already stamped on produced artifacts) into individual tool-call audit records, so every external action is attributable to the run that caused it.

- `agent/audit_trail.py`: new `current_attribution()`; `append()` stamps `metadata.attribution` plus a top-level `attribution` field when the process runs under an attributed subagent. Existing metadata is preserved; fully inert without the env marker (internal bridge var, not user config).
- `tests/test_audit_trail_attribution.py`: 3 unit tests (default off / stamped / metadata preserved).

## Validation
- ruff check + format ✓; targeted tests 3/3 ✓

### DIFF
--- a/agent/audit_trail.py
+++ b/agent/audit_trail.py
@@ -152,10 +152,32 @@ def sanitize_metadata(metadata: Optional[Dict[str, Any]]) -> Dict[str, Any]:
     return cleaned
 
 
+def current_attribution() -> Optional[str]:
+    """Return the subagent attribution marker for this process, if any.
+
+    evo-2026-08-26-03: propagate the ``HERMES-SUBAGENT-ATTRIBUTION`` marker
+    beyond produced artifacts into individual tool-call audit records, so
+    every external action is attributable to the run that caused it.
+    """
+    return os.environ.get("HERMES-SUBAGENT-ATTRIBUTION") or None
+
+
 def append(record: dict, *, path: Path | None = None) -> Optional[dict]:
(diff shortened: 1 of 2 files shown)


Lexus2016:
Blocked at integration review: dead wiring — `current_attribution()` reads env var `HERMES-SUBAGENT-ATTRIBUTION`, but NOTHING in the codebase ever sets it (verified by grep across all *.py on main). The attribution stamp exists only as prompt-block/artifact-header text built by `tools/delegation_attribution.build_attribution_stamp()`; tests pass only because they set the env var via monkeypatch, so the feature would be permanently inert in production.

To fix (re-open a PR once wired): set the stamp into the child's environment at delegation time — e.g. in `tools/delegate_tool.py` around each child's run, `os.environ['HERMES-SUBAGENT-ATTRIBUTION'] = _attribution_stamp` with save/restore (children are in-process threads sharing os.environ, so guard against cross-thread leakage), or thread it via the existing ContextVar used for HERMES_SESSION_ID and read that in `current_attribution()`. Definition of done: an audit record written during a real delegated subagent tool call carries a non-null `attribution` matching the child's stamp — verified end-to-end, not via monkeypatch.

Issue #3210, opened 2026-08-26
[IMPROVEMENT] Convert experiential memory into an evidence-grounded, trigger-monitored skill library


Lexus2016:
### Problem Statement
Memory failures are a recurring top failure bucket in Hermes pipeline metrics. Hermes already stores memories and already has a `SKILL.md` corpus plus a `SkillCrystallizer` that can turn successful traces into provisional skill candidates. What is missing is the operational link: experiential memory mostly sits as passive context the model must re-reason over, rather than being promoted into skills with explicit trigger conditions that are monitored at runtime and injected when the state matches.

### Proposed Solution
Evolve the memory-to-skill path into an **evidence-grounded co-evolution governance** layer. For each validated memory/trace, produce a skill entry with an explicit trigger function μ(state, task-kind). At runtime, the agent monitors the current state and, when a trigger fires, retrieves the crystallized skill into context to guide action instead of re-deriving the workflow from scratch.

Concrete steps:
1. **Evidence bar** — only promote memories that pass the existing `meets_evidence_bar` gates (reusability ≥ 0.6, ≥ 2 distinct tools, ≥ 3 actions, verified success). This guardrail already exists in `evolution/lib/skill_crystallizer.py` and must remain hard.
2. **Trigger extraction** — when crystallizing a skill, extract and store the trigger conditions under which it should fire (task kind, tool constellation, error class, user intent signal). Store them in the skill frontmatter so they are inspectable.
3. **Runtime trigger monitor** — add a lightweight monitor that, before each model turn, scores current state against stored trigger conditions and retrieves matching skills when the score exceeds a calibrated threshold.
4. **Co-evolution loop** — when a triggered skill is used, record outcome; demote or revise skills whose triggered use repeatedly fails, so the library improves rather than fossilizes.

### Value Proposition
- **Impact**: High (0.8)
- **Effort**: Medium (0.5)
- **Priority Score**: 1.28

### Research Evidence
- [From Memory to Skills: Evidence-Grounded Co-Evolution Governance for Long-Horizon LLM Agents](https://arxiv.org/abs/2607.16621)

### Implementation Plan
1. Extend `evolution/lib/skill_crystallizer.py` to require and validate trigger metadata in skill frontmatter.
2. Add a trigger matcher/scorer (state + skill trigger → relevance score) in a new small module under `evolution/lib/`.
[...]

PR #3211, opened 2026-08-26
feat: trigger metadata + matcher for crystallized skills — increment 1 of #3210

Lexus2016:
First coherent slice of #3210 (evidence-grounded, trigger-monitored skill library).

What lands:
- `evolution/lib/trigger_matcher.py`: deterministic trigger extraction from traces, frontmatter validation/parse/render, runtime state->trigger scoring (`score_trigger`, `best_matches`).
- `evolution/lib/skill_crystallizer.py`: every crystallized skill now carries a validated `triggers:` block in SKILL.md frontmatter; `validate_candidate` rejects malformed trigger blocks (back-compat: candidates without triggers stay valid).
- Tests: `tests/evolution/test_trigger_matcher.py` (20 tests incl. existing suite green; full `tests/evolution` shard 431 passed locally).

Note: 308 changed lines — above the 200-line autonomous self-merge cap; the merge gate should hold this for human review.

Deferred (next increment):
- Runtime trigger monitor wired into the agent turn loop (prompt-cache-safe injection).
- Co-evolution loop: outcome logging + automatic demotion of repeatedly failing triggered skills.

### DIFF
--- a/evolution/lib/skill_crystallizer.py
+++ b/evolution/lib/skill_crystallizer.py
@@ -18,6 +18,11 @@ from typing import Any, Dict, List, Optional, Sequence, Tuple
 
 logger = logging.getLogger(__name__)
 
+try:
+    from evolution.lib import trigger_matcher as _tm
+except ImportError:  # pragma: no cover - direct-path import fallback
+    import trigger_matcher as _tm  # type: ignore[no-redef]
+
 __all__ = [
     "CrystallizedSkillCandidate",
     "SkillCrystallizer",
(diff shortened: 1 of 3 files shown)


Lexus2016:
Integration review: blocked in-cycle. The new trigger_matcher module adds real value to skill_crystallizer frontmatter (extract/validate/render/parse are wired), but  and  have zero production call sites in this increment — they are referenced only by their own module and tests. Per the evolution integration dead-code gate, autonomous merge is not permitted. The PR is also above the 200-line self-merge cap, which the author already flagged. Please rework so the scoring surface ships with its runtime consumer (the deferred trigger monitor wiring), not as dead code.

PR #3216, opened 2026-08-26
fix(evolution): inject introspection pre-run digest via cron preflight (Closes #115)

Da-Mikey:
The evolution-introspection skill's process step 1 says to work from a "pre-computed introspection digest" that the cron pre-run script injects into the prompt under "## Introspection Digest". That section never arrives: the pre-run script (evolution_access_gate.sh) only checks GitHub write access and emits the wake-gate JSON. It does not run the session-signal extractor, so the agent must re-derive the digest itself every cycle.

- Frequency: digest absent in 5+ consecutive cycles (Aug 20–25); each cycle's own report notes "the pre-computed digest was missing from the prompt again"
- Tool/area involved: cron pre-run script wiring + evolution-introspection skill process step 1
- Failure shape: prompt contains only the access-gate line; the expected digest section is never present; the agent falls back to running the extractor manually via terminal

### Impact on real tasks
Each weekly cycle burns LLM budget and terminal calls re-extracting session signals that should be pre-computed. The manual extraction path is exactly the pattern that caused the documented July 19 loop-guard kill of this job (agent stuck chaining terminal calls in unattended cron). The skill's own pitfall section warns about this loop; the missing digest actively recreates the conditions for it every run.

### Proposed direction
Wire the extractor into the pre-run: after the access gate passes, run the session-signal extractor and either (a) inject its compact anonymized JSON into the prompt context, or (b) write it to a fixed path the skill tells the agent to read. Update the skill's process step to fall back gracefully ("if no digest is present, run the extractor ONCE, then proceed") instead of expecting an artifact that is never produced.

### Value
- Impact: 0.6 (every cycle: wasted budget + recurring loop-guard risk in unattended cron)
- Effort: 0.1 (small pre-run wiring change)
- Priority Score: 12.0

---

## Implementation (round 1, 2026-08-26)

- scripts/introspection_extract.py gains --out=FILE (deterministic digest to
  a fixed path).
- New scripts/introspection_preflight.py runs the extractor before the LLM
  wakes and always emits the wake signal (digest injection, not suppression).
- cron/evolution/introspection.yaml wires the preflight as its pre-run
  script; the introspection skill now reads the pre-written
  introspection/{date}-digest.json first and only re-runs the extractor when
  it is missing.
- Selected by analysis phase 2-github-aware-merged (2026-08-26).

Closes #115


Da-Mikey:
Closing: empty placeholder (write_text bug); issue #115 rejected by phase-3. Superseded by 3212/3213/3214.

PR #3222, opened 2026-08-27
fix(process): accept 'stop' as an alias for 'kill' (Closes #3221)

### LINKED ISSUE #3221: [FIX] process tool: model repeatedly uses invalid action='stop' (5 sessions/7d)
The agent repeatedly invokes the `process` tool with `action='stop'`, which is not a valid action. The `process` tool schema allows `list`, `poll`, `log`, `wait`, `kill`, `write`, `submit`, `close`; `stop` is not among them. Each invalid call returns an argument-validation error and wastes a turn before the agent retries with the correct verb.

- Frequency: 5 occurrences across 5 distinct sessions in the last 7-day window (rate 0.0051/session); 4 share a single timestamp (broadcast-duplicate artifact), plus 1 cron session — at least 2 independent contexts.
- Tool/area involved: `process` tool argument schema / action enum.
- Failure shape: `Invalid arguments for 'process': 'action'='stop' is not one of the allowed values: ...` — the model consistently hallucinates the natural verb `stop`.

### Impact on real tasks
The `process` tool manages background task lifecycle (poll, wait, kill). When the agent needs to terminate a background process, using `stop` fails immediately. The error message does list allowed values, so recovery is fast, but the repeated same-verb hallucination wastes a turn every time and indicates the schema/docs do not align with the model's natural usage.

### Proposed direction
Add `stop` as an accepted alias for the lifecycle-terminating action in the `process` tool schema, mapping it to the existing `kill` behavior. Alternatively, update the tool description to explicitly mention "use `kill`/`close`, not `stop`" and improve the argument-error message to suggest the correct verb. The fix should be one-line in the schema enum plus handler dispatch.

### Value
- Impact: 0.4
- Effort: 0.1
- Priority Score: 0.768

Lexus2016:
Closes #3221

Models frequently invoke the process tool with action='stop' when they want to terminate a background process. The schema only allowed 'kill', so every 'stop' call produced an argument-validation error and wasted a turn before retrying.

This PR adds 'stop' to the process tool schema enum and dispatches it to the same handler as 'kill'.

Checks (local):
- Targeted tests: pytest tests/tools/test_process_schema_confusion.py tests/tools/test_process_registry.py (119 passed, 4 skipped)
- Blocking lint: ruff check . passed
- Pre-PR runner: python scripts/evolution_pre_pr_test_runner.py --changed-files tools/process_registry.py,tests/tools/test_process_schema_confusion.py passed
- Diff size: 28 changed lines (≤ 200 self-merge cap)

### DIFF
--- a/tests/tools/test_process_schema_confusion.py
+++ b/tests/tools/test_process_schema_confusion.py
@@ -21,7 +21,6 @@ class TestSuggestProcessAction:
         ("run", "list"),
         ("status", "poll"),
         ("check", "poll"),
-        ("stop", "kill"),
         ("tail", "log"),
         ("read", "log"),
         ("send", "submit"),
@@ -40,7 +39,7 @@ class TestSuggestProcessAction:
 
     def test_valid_action_returns_none(self):
         """Valid actions don't need a suggestion."""
-        for action in ["list", "poll", "log", "wait", "kill", "write", "submit", "close"]:
+        for action in ["list", "poll", "log", "wait", "kill", "stop", "write", "submit", "close"]:
             # Valid actions may or may not return a suggestion — the
             # handler only calls _suggest_process_action for INVALID actions
             pass  # no assertion needed, just verify no crash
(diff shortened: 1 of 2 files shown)


Da-Mikey:

PR #3249, opened 2026-08-27
feat(evolution): checked lowering for declared tool inputs (Closes #77)

Da-Mikey:
Automated evolution PR for #77 (checked lowering, S1+S2 rework r3).

Implements the rework brief's definition of done:
- **Real enforcement path:** `find_cap_violations()` is wired into `find_violations()`, which IS the call site used by the registration-time pre-flight (`scripts/register_evolution_cron.py` calls `lint_mod.find_violations(...)`) — the cap gate now runs where the pipeline actually gates skills, not only under `lint_repo()`.
- **Lazy import (`_cap_checker`):** fixes the r1/r2 `ModuleNotFoundError` that silently disabled the whole skill→toolset pre-flight under importlib file-spec loading (only `repo_root` on sys.path). Degrades to a documented no-op if the sibling module is unreachable; other lint rules still run.
- **Opt-in declared inputs:** a skill is only checked if it declares `declared_inputs:` in frontmatter — zero false positives on the existing corpus.
- **Tests:** `tests/scripts/test_evolution_tool_cap_check.py` — 17 tests, all passing locally; `ruff check` clean; `ruff format --check` clean.

Out of scope (per the 08-19 decomposition): S3 bounded execution (live tool path) needs its own design review; core model-tool surface untouched.

### DIFF
--- a/scripts/evolution_skill_lint.py
+++ b/scripts/evolution_skill_lint.py
@@ -37,6 +37,27 @@ are the pure core (unit-tested with synthetic diff stats); ``lint_skill_edit_bud
 is the git-diff IO boundary, run via ``--skill-edit-budget`` (not part of the
 default ``lint_repo()`` static check, since it is inherently relative to a base
 ref — see ``main()``).
+
+Part 3 — checked lowering for declared tool inputs (#77, SkillEffect, S2)
+-------------------------------------------------------------------------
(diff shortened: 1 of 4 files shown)


Lexus2016:
Rebase-merged into `main` via commit c7aa19b8b2 with author credit preserved.

PR #3250, opened 2026-08-27
feat(memory): per-tier memory-tool injection dosage (Closes #75)

Da-Mikey:
Automated evolution PR for #75 (rework r4 — memory dosage slices 1+2).

**Rework decision (documented):** the previous review brief demanded fixing a
config key-path mismatch (`memory_injection` read at top level in
`agent/agent_init.py` vs defined under `agent` in `config_defaults.py`). That
target is MOOT on current main: `git log -S memory_injection` shows the
config-based dosage feature **never landed on main** — the key does not exist
anywhere in `agent/` or `hermes_cli/`. There is no config path to fix.

**This PR implements the issue's decomposition slices 1+2 at the real
injection site** (`inject_memory_provider_tools`, called at session start from
`agent/agent_init.py:1978`):
- Per-tier injection profile: how many memory-provider tool schemas get
  appended to the tool surface is capped by model capability (mini/flash/small
  = 3, default = 6, pro/large = 8).
- Tiered caps rather than one global number; operator override via
  `EVOLUTION_MEMORY_TOOL_CAP` (internal escape hatch, precedent:
  `SKILL_TRUSTED_SOURCES` in tools/skills_tool.py).
- Prompt-cache safe: cap is computed once at session start, does not mutate
  mid-conversation.

Validation: `ruff check` clean, `ruff format --check` clean,
`tests/agent/test_memory_provider.py` 67 passed.

### DIFF
--- a/agent/memory_manager.py
+++ b/agent/memory_manager.py
@@ -27,6 +27,7 @@ from __future__ import annotations
 
 import json
 import logging
+import os
 import re
 import inspect
 import threading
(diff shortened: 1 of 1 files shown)


Lexus2016:
Rebase-merged into `main` via commit 9d343681d4 with author credit preserved.

PR #3251, opened 2026-08-27
fix(skills): reviewed-allowlist escape hatch for SkillTrojan false positives (Closes #120)

Da-Mikey:
Automated evolution PR for #120.

**Problem:** the SkillTrojan composition defense false-positives on
cross-skill URL-fragment references — it blocks `skill_view` on legitimately
pipelined skills and, because cron jobs attach those skills, skips unrelated
cron jobs (the evolution pipeline was starved). The injection block had an
escape hatch (`SKILL_TRUSTED_SOURCES`) but the composition trace had none.

**Fix:** operator-controlled, name-based reviewed-allowlist escape hatch:
- `SKILL_REVIEWED_ALLOWLIST` (comma-separated skill names) marks a skill as
  explicitly reviewed by the operator.
- Reviewed skills bypass the injection block (consistent with the existing
  trusted-sources path) and the composition trace — the same trust decision
  the operator already makes with `SKILL_TRUSTED_SOURCES`, applied to the
  check that had no escape hatch.
- Blocked-skill error message now points operators at the new escape hatch.
- Default behavior unchanged: allowlist empty → identical guard behavior.

Validation: `ruff check` + `ruff format --check` clean;
`tests/tools/test_skills_tool.py` + `test_skill_composition_tracer.py` 41 passed.

### DIFF
--- a/tools/skills_tool.py
+++ b/tools/skills_tool.py
@@ -858,7 +858,9 @@ def _find_all_skills(*, skip_disabled: bool = False) -> List[Dict[str, Any]]:
             skill_dir = skill_md.parent
 
             try:
-                content = skill_md.read_text(encoding="utf-8-sig", errors="replace")[:4000]
+                content = skill_md.read_text(encoding="utf-8-sig", errors="replace")[
+                    :4000
+                ]
                 frontmatter, body = _parse_frontmatter(content)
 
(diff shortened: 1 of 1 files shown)


Lexus2016:
Rebase-merged into `main` via commit 65fd629404 with author credit preserved.

PR #3252, opened 2026-08-27
fix(memory): contradiction detector generic-token stopwords (Closes #121)

Da-Mikey:
Automated evolution PR for #121.

**Problem:** the contradiction detector over-fires on unrelated observations
that share generic context tokens — 903 warnings in one day (1,090 per the
analysis window), risking wrong memory rewrites and flooding logs.

**Root cause:** the generic-token stopword list was too thin. Tokens like
`system`, `pipeline`, `memory`, `data`, `user`, `agent`, `time`, `job`,
`task`, `session`, `log`, `status`, `error` co-occur across unrelated system
state / pipeline log lines, so unrelated observations pass the shared-token
bar and get flagged as contradictions.

**Fix:** 19 new polarity-neutral stopwords. The shared-token bar stays at 2 —
subject+claim pairs like "nas healthy" vs "nas not healthy" are *genuine*
contradictions and must keep firing (the analysis-stage draft raised the bar
to 3; tests showed that suppresses real detection, so this PR deliberately
keeps the bar and fixes the noise at the token level instead).

Validation: `tests/agent/test_memory_contradiction.py` +
`test_memory_conflicts.py` 47 passed; `ruff check` + `ruff format --check` clean.

### DIFF
--- a/agent/memory_contradiction.py
+++ b/agent/memory_contradiction.py
@@ -294,12 +294,15 @@ def detect_contradictions(
         if stored_negated == obs_negated:
             continue
         shared = obs_tokens & _content_tokens(stored_text)
-        # Require at least THREE shared content tokens (#121). A single token
-        # is usually a common predicate; two still fires on unrelated
-        # observations sharing a generic subject + common predicate (the
-        # over-fire that produced 1,090 warnings in 3 days). Three shared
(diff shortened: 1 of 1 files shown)


Lexus2016:
Rebase-merged into `main` via commit 3ac7516d4e with author credit preserved.

Issue #3093, opened 2026-08-22
[evolution] Toolset mismatch blocks practical task completion


Lexus2016:
## Problem pattern

Agent sessions are launched with only web-search/extract tools. When a user authorizes and requests a local file write or shell command, the agent cannot act and must refuse. This violates the "Finishing the job" principle that asks for a working artifact backed by real tool output.

## Indicators

- Assistant reports missing file-write or terminal tool.
- Task explicitly requires filesystem action but no filesystem tool is exposed.
- Multiple recent sessions hit the same wall for the same capability-class request.

## Proposed fixes

1. Ensure sessions that may receive build/run/verify tasks expose filesystem and terminal tools.
2. Add a pre-flight capability check that declines or escalates requests requiring tools the session does not have.
3. If toolsets are intentionally restricted, surface that limitation to the user before they issue action-oriented requests.

## Privacy note

This issue is based on abstracted session analysis; no raw user text or PII is included.

PR #3295, opened 2026-08-28
feat(delegate): auto-add web toolset to subagents whose goal needs it (Closes #126)

Da-Mikey:
Automated evolution PR for issue #126 — delegated research/sync subagents arrived with no host web fallback when their only provisioned web path (an MCP toolset such as firecrawl) was exhausted, and announced the gap mid-run instead of completing. Mirrors the existing terminal (#1369) and file (#3093) pre-dispatch compatibility checks: static verb heuristic on goal+context text, bounded by the parent-intersection so a subagent never gains tools its parent lacks.

Closes #126

Tests: 20 targeted delegate tests + full delegate suite (364) green locally; ruff clean.

### DIFF
--- a/tests/tools/test_delegate_web_compat.py
+++ b/tests/tools/test_delegate_web_compat.py
@@ -1,0 +1,76 @@
+"""Tests for the pre-dispatch web-toolset compatibility check (#126).
+
+Verifies that:
+1. ``_goal_needs_web`` detects web-dependent verbs in goal/context text.
+2. The auto-add logic adds ``web`` when the goal needs it but the resolved
+   toolset omits it — but only when the parent can provide it (no widening).
+3. Research-style goals that previously shipped a child with no host web
+   fallback now resolve with ``web`` present.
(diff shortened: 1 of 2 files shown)


Da-Mikey:
Evolution pipeline status — CI is red for a repo-wide infra reason, NOT this PR.

`scripts/run_tests_parallel.py` (broken by the v2026.8.27 sync, PR #3274) uses `temproot` in a cleanup `finally` block without ever defining it, so every test slice crashes with `NameError: name 'temproot' is not defined` on every open PR (verified in the failing job logs; deterministic — a re-run fails identically, so no re-run was burned).

Code review passed on this PR: dead-code and judgement checks against live main (`_goal_needs_web` has a real call site in `_build_child_agent` next to the #1369/#3093 compat checks; parent-bounded, no widening; `web` is a valid toolset in `DEFAULT_TOOLSETS`). Will re-verify CI and hand off once the runner fix lands.

PR #3398, opened 2026-08-31
feat(memory): TEPA revocable memory — explicit validity state on episodic events (Closes #154)

Da-Mikey:
Automated evolution PR for issue #154 — **TEPA: revocable evidence-memory**.

## What & why
Long-term memory accumulates stale beliefs (renumbered IPs, superseded creds, dead crons) that stay retrievable and pollute the prompt. Current mitigations are cleanup heuristics (overwrite, rank-down, age-out) that lose the *reason* a belief was replaced. This change models memory observations as keyed precedents with **validity as explicit state**: an active precedent is **revoked** when fresh evidence contradicts it, rather than overwritten or left to age out.

## Implementation (issue's 3-step plan)
1. **Explicit validity field on the entry schema** — `MemoryEvent` gains `validity` (`active`/`revoked`), `revoked_at`, `revoked_by` (superseding evidence id), `revocation_reason`. Serialization is backward-compatible: legacy dicts without the new keys load as `active`.
2. **Revocation on contradictory evidence in the write path** — `MemoryManager.score_memories` (the episodic write path) now revokes each stored event flagged by the existing contradiction detector instead of leaving both sides at full weight. Non-fatal by design: a revocation failure logs and is skipped; the turn sync never fails.
3. **Revocation queryable** — `EpisodicMemoryStore.revoked_events()` returns the audit trail (what/when/by-whom/why); every retrieval surface (`text_search`, time-range, category, importance, temporal proximity) gains `include_revoked` and **excludes revoked entries by default**, so a stale claim never resurfaces in later prompts.

**Revoke ≠ delete**: the record is kept (coordinates with the graveyard/purge flow); `deduplicate` only ever merges active events, so a revoked entry can never be resurrected; the contradiction scan now runs over active events only, so dead memories stop generating noise.

## Files
- `agent/memory_importance.py` — schema fields, `revoke()`, `revoked_events()`, `active_events()`, `_events()` filter, `include_revoked` on all retrieval + dedup
- `agent/memory_manager.py` — write-path revocation (`_revoke_contradicted`), contradiction scan over active events
- `tests/agent/test_memory_validity.py` — 16 new tests (schema, serialization/legacy compat, revocation semantics, retrieval exclusion, dedup safety, save/load, write-path wiring, non-fatal failure, no-noise-after-revocation)

## Validation (local, matching CI)
- `ruff check` clean, `ruff format --check` clean
- `pytest tests/agent/test_memory_validity.py tests/agent/test_memory_importance.py tests/agent/test_memory_contradiction.py` → **81 passed** (16 new + 65 existing, no regressions)

### DIFF
--- a/agent/memory_importance.py
+++ b/agent/memory_importance.py
@@ -13,6 +13,11 @@ A self-contained module (stdlib only) that:
 4. Deduplicates events via Jaccard similarity on tokenized text, merging
    groups while keeping the highest-importance event and combining tags.
 5. Persists the whole store to / loads it from a JSON file.
+6. Tracks explicit validity state (TEPA, #154): a stored event
+   contradicted by fresh evidence is *revoked* — marked with validity
+   state, superseding id and reason — rather than overwritten, and
+   revoked entries are excluded from normal retrieval while staying
+   queryable for audits.
 
 Import-safe: importing this module has no side effects and requires no
 third-party packages.  Designed to be unit-testable with only stdlib +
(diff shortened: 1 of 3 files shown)


Lexus2016:
Merged into main in commit 2ada986657. Fully verified with 94 tests passing and zero regressions.

PR #162, opened 2026-06-13
docs/chore(fork): MIT license, fork metadata, English EVOLUTION_README, pytest marks

Lexus2016:
Fork-correctness fixes from AUDIT_REPORT.md (findings #1, #2, #11, #12, #14):

- **License (#1):** README.md badge + text and EVOLUTION_README.md Apache 2.0 → **MIT** (LICENSE, pyproject, package.json are all MIT).
- **Fork metadata (#2):** package.json `version` 1.0.0 → 0.16.0; `repository`/`bugs`/`homepage` URLs NousResearch/Hermes-Agent → Lexus2016/hermes-agent-evolution.
- **EVOLUTION_README → English (#14 + owner request):** fully translated UA → EN (project docs are English). Also corrected the schedule table to the real cron (Mon/Wed/Fri upstream + introspection/integration/watchdog), switched setup to `scripts/register_evolution_cron.py`, documented the bounded/PR-level upstream sync + version marker, and linked SECURITY_EVOLUTION.md.
- **pytest marks (#11):** registered `xdist_group` + `ssh` in pyproject (silences PytestUnknownMarkWarning / unblocks `--strict-markers`).
- **Test command (#12):** EVOLUTION_README now points to `python scripts/run_tests_parallel.py`.

Note: audit finding **#16** (`release.py:1303` `iuyup`) is a **false positive** — it is an AUTHOR_MAP entry (email → GitHub handle), not a credential; left untouched.

Docs/metadata only — no code changes.

PR #3399, opened 2026-08-31
fix(evolution): watchdog audit tolerates nested effort_budget shapes (Closes #168)

Da-Mikey:
Automated evolution PR for issue #168: meta-skill trace crashed with TypeError when effort_budget was a nested dict (08-28/08-29 reports). Now reads scalar explicitly + loud shape check at producer and consumer.

### DIFF
--- a/scripts/evolution_analysis_audit.py
+++ b/scripts/evolution_analysis_audit.py
@@ -58,6 +58,71 @@ def _num(x: Any) -> Optional[float]:
     return None
 
 
+# Keys under which a nested ``effort_budget`` dict may carry its scalar value.
+# Observed shapes (analysis reports): ``{"limit": 3.0, ...}`` (2026-08-29),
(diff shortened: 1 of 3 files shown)


Da-Mikey:
CI state after rework (2026-09-01): the original regression (test_missing_access_gate_fails_closed — import crash in isolated installs) is FIXED: all 6 gate tests pass locally and the gate slices are green. The remaining red test slices are a PRE-EXISTING main/CI issue, not this PR: the same 8-slice failure pattern appears on unrelated dependabot PR #3408 (pure GH Action version bump), and the failing tests (test_config_drift, test_bedrock_integration, test_prompt_size, test_patch_*, test_process_registry, test_no_shadowed_test_definitions) all PASS locally on both this branch and lexus/main. This is the 'MAIN RED — 7 tests fail on every open PR' class documented in issue #162 (which was closed accepted but the class persists with a partially different test set). No evolution-code path is implicated.

Lexus2016:
Merged into main in commit 085b12f033. All tests passing.

Issue #701, opened 2026-07-03
[IMPROVEMENT] Detect and alert when evolution cron jobs complete with zero tool calls


dimokru:
### Problem Statement
Hermes cron jobs currently report `cron_complete` even when the agent session made zero tool calls, which happens whenever a required toolset is broken or missing in the runtime environment. Recent sessions show the evolution pipeline continuing to schedule downstream stages while earlier stages silently no-op, because completion is binary rather than outcome-based.

### Proposed Solution
Add a "minimum tool-call / output artifact" health check inside `evolution_funnel.py` (or the no-agent watchdog): a stage that produced no sidecar change and zero tool calls should be flagged as `stage_failed`, not counted as `completed`.

### Value Proposition
- **Impact**: High (0.8)
- **Effort**: Low (0.3)
- **Priority Score**: 1.41

### Research Evidence
- Source: `~/.hermes/profiles/user1/evolution/introspection/2026-07-02.json`
- Source: `scripts/evolution_funnel.py`
- Source: `cron/evolution/*.yaml`

### Implementation Plan
1. Read stage sidecar and session summary to count tool calls.
2. Define a per-stage minimum threshold (e.g., at least one tool call and a changed sidecar file).
3. Emit `stage_failed` with a diagnostic label when the threshold is missed.
4. Wire the result into `evolution_funnel.py` so downstream stages are not scheduled.
5. Add a unit test using a synthetic zero-tool-call session summary.

### Success Criteria
- [ ] A cron job with zero tool calls is marked `stage_failed`, not `cron_complete`.
- [ ] Downstream stages are not scheduled for a failed stage.
- [ ] Existing healthy sessions still pass the check.

PR #3418, opened 2026-09-09
fix(cron): hoist delivery/refusal tail above partial-load tear point (part 2)

Da-Mikey:
## Why

`32e90b53` hoisted `_LAST_RUN_TOOL_CALLS` to stop NameError on partial module loads, but the cron delivery/refusal tail still lives at the **bottom** of `cron/scheduler.py`. A gateway importing `cron.scheduler` while `hermes update` replaces the file on disk can truncate the module before those definitions, so run/delivery paths die with:

```
NameError: name '_maybe_cron_refusal_recovery' is not defined
```

Observed **daily since Sep 4** on this install despite the #701 fix: Devyn Daily Cosplay Email, Hestia Daily Audit, family-calendar-daily, Iris Proactive Check-In, FutureTools Newsletter Scanner — jobs crash post-run and records are lost. (The error reads like a code bug, so it gets debugged as one — same trap documented at `_summarize_cron_failure_for_delivery` #95294.)

## What

Move the whole late-defined tail block — `strip_reasoning_for_delivery`, `_count_tool_calls`, `_maybe_cron_refusal_recovery`, `_failure_digest_enabled`, `_load_cron_config`, `build_cron_failure_digest`, `_summarize_job_result`, `_resolve_delivery_verbosity`, `_apply_delivery_verbosity` — up to sit immediately after the import preamble (same pattern as the #701 fix). Pure transposition, zero logic changes.

Also adds `Any` to the top-level `typing` import (the moved functions annotate with it; it was previously resolved by a later import).

## Verification

- `py_compile` clean
- module import clean
- functional probes on all moved helpers (verbosity resolution, tool-call counting, strip, refusal-recovery passthrough)
- `tests/cron/test_cron_refusal_nudge.py`: **7 passed**

Note: `git diff` shows ~17k lines because the diff engine can't align a 370-line block transposition — verify with `git diff --color-moved` / `--ignore-space-at-eol`; the actual content delta is 7 lines (verified programmatically: every old line except the modified typing import exists verbatim in the new file).


Lexus2016:
Thanks — the remaining hoist is real, but this branch conflicts with `bdabcb6827` which already moved `_maybe_cron_refusal_recovery`.

That commit left the sibling path open: recovery (now at the top) still calls `_count_tool_calls`, which lived below `if __name__`, so a prefix/truncated import that reaches `_run_job_impl` still NameErrors. Delivery helpers had the same tear point.

Salvaged onto current `main` as #3419 (function bodies unchanged; `_count_tool_calls` sits above recovery). Please don’t merge this PR — a merge would double-hoist recovery.

Lexus2016:
Superseded by #3419, now merged.

This branch conflicted with `bdabcb6827` (recovery already hoisted). The remaining sibling path (`_count_tool_calls` + delivery helpers) landed on `main` via #3419.

